Issue #1, opened 2016-03-11
Basic, master styles


kookster:
ripped and translated from www

PR #3, opened 2016-03-15
Global styles

cavis:
Port styles and layout from the www-app:

- Grab the raw css (from the source stylus files)
- Use `ViewEncapsulation.None` in the root app component, so styles cascade to everything
- Convert header/footer styles to be scoped


cavis:
This covers #1 and #2.

chrisrhoden:
@cavis Is there a benefit to using `ViewEncapsulation.None` rather than using a <style> tag for global styles?

cavis:
It should be functionally equivalent (it writes those as unscoped `<style>` tags).  Which was really the only way to make those styles apply to child components of the root app.

My thought was that it's the "Angular 2 way" to let your components load all the styles, rather than putting them in `index.html` yourself.  But I'm curious how others out there deal with global styles.  I couldn't find any great examples.

kookster:
I also have not yet seen a great example of how to do this, and both options make sense to me.

I do think we should go back to the standard file naming for `app.ts` => `app.component.ts`

src/app/footer/footer.ts
```diff
@@ -0,0 +1,56 @@
+import {Component} from 'angular2/core';
+import {Location} from 'angular2/router';
+
+@Component({
+  selector: 'publish-footer',
+  styleUrls: ['app/footer/footer.css'],
+  template: `
```
kookster: Pretty big template - when does it justify a separate file?

chrisrhoden: My 2c: At this point I am starting to think that inline is almost always better with the caveat that current tools don't give you syntax highlighting (which I expect to change).

kookster: meant it as an honest question. I agree on the highlighting - seems like from an editing standpoint, that is a trade-off, but it is nicer to have inline where the inputs, directives, and whatnot are also right in front of you.

src/app/shared/spinner/spinner.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  @Input() public set spinning(value: boolean) {
+    if (!value) {
+      this.cancelTimeout();
+      this.isSpinning = false;
+    }
+    if (this.currentTimeout) {
+      return;
+    }
+    this.currentTimeout = setTimeout(() => {
```
kookster: I have only read about these, so this is just a question not a critique, but is there any advantage to using an rxjs future scheduler?

https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/schedulers/scheduler.md#rxschedulerprototypeschedulefuturestate-duetime-action

kookster:
lgtm :+1:

src/app/shared/spinner/spinner.component.ts
cavis: Looks pretty much equivalent to me.  The rxjs syntax might be slightly cleaner - I'll probably try to use more of it going forward.

FWIW, I'm not sure that delay/timeout stuff will actually be useful for us, so it may end up just getting nuked at some point.  (After we build a couple different pages with spinners on them).

PR #4, opened 2016-03-17
`Env` configuration

kookster:
Generate a `src/util/env.ts` file from the contents of `.env` using a gulp task.

Defines an `Env` class like this:
`.env`
```
FOO=bar
```

`src/util/env.ts`
```
// GENERATED FILE, DO NOT EDIT OR CHECK IN
export class Env {
  public static FOO = 'bar';
}
```


kookster:
I am gonna stop messing with this, should be good now.

cavis:
I like the spirit of this.  And should leave things open to somehow "live-reload" changes to `.env`.

My only question is: how do we deal with default and/or ci-testing values?  Would we just add another `.env.testing` file?  Or expect the CI-env to export all the ENVs it would need to run?

kookster:
that is a good question. I hadn't thought that far ahead I guess. Maybe in the CI scripts, it could run the gulp task, and the `.env` file could either be in the src, or maybe the ci script could add it based on config in the ci system/script?

cavis:
My first thought was to have a `.env.defaults` added to git, which uses "sane defaults" that also happen to make the tests pass if a `.env` is not present.

Or maybe we just write tests that don't require specific configs.

kookster:
I think you are right, we can do that, and for the upload bucket, we can make a test bucket and credentials that delete all files after 24 hrs. I will work on that.

cavis:
Merging this.  Will iterate in future PRs.

PR #5, opened 2016-03-17
Basic File uploads

kookster:
* EvaporateJS
* Test page for uploads
* upload service
* file select directive
* display uploading file name, progress %
* cancel button


app/upload/evaporate.d.ts
```diff
@@ -0,0 +1,9 @@
+declare class Evaporate {
```
cavis: Will this definition eventually move into evaporate itself?

kookster: yeah, I have submitted it as a PR:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/8742

app/upload/upload-file-select.directive.ts
```diff
@@ -0,0 +1,22 @@
+import {Directive, ElementRef, EventEmitter} from 'angular2/core';
+import {UploadService} from './upload.service';
+
+@Directive({
+  selector: '[upload-file]',
+  inputs: ['uploadService'],
```
cavis: General question: what does this `inputs` do?

app/upload/upload.component.ts
```diff
@@ -0,0 +1,20 @@
...
+import {HTTP_PROVIDERS} from 'angular2/http';
+import {FileUpload} from './file-upload.component';
+
+@Component({
+  template: `
+    <input type="file" upload-file />
+    <file-upload *ngFor="#upload of uploadService.uploads" [upload]="upload">
+    </file-upload>
+  `,
+  providers: [HTTP_PROVIDERS, UploadService, MimeTypeService],
```
cavis: Where are `HTTP_PROVIDERS` and `MimeTypeService` used?

app/upload/upload.service.ts
```diff
@@ -0,0 +1,86 @@
...
+      bucket: this.bucketName,
+      cloudfront: true
+    });
+  }
+
+  addFile(file:File, contentType?:string) : Upload {
+    let ct = contentType || this.mimeTypeService.lookupFileMimetype(file).full();
+    let upload = new Upload(file, ct, this.evaporate);
+    this.uploads.push(upload);
+    // upload.upload();
```
cavis: No longer needed?

app/upload/upload-file-select.directive.ts
kookster: That may be an old way to specify an @Input, that was in the docs when I looked before, but it is gone now - I'll try updating it to the annotation style.

config/systemjs.config.js
```diff
@@ -108,21 +112,17 @@ System.config({
...
     },
     "npm:angular2@2.0.0-beta.12": {
-      "crypto": "github:jspm/nodelibs-crypto@0.1.0",
-      "es6-shim": "npm:es6-shim@0.35.0",
-      "process": "github:jspm/nodelibs-process@0.1.2",
-      "reflect-metadata": "npm:reflect-metadata@0.1.2",
+      "reflect-metadata": "npm:reflect-metadata@0.1.3",
       "rxjs": "npm:rxjs@5.0.0-beta.2",
-      "zone.js": "npm:zone.js@0.6.6"
+      "zone.js": "npm:zone.js@0.5.15"
```
cavis: Hmm... this downgrade is a bit worrying.  Is hot-reloading still working for you?  I thought I needed `0.6.6` to get it working.  But I'll pull down this branch and re-check.

kookster: agreed, that is odd, and I didn't do that explicitly, let me look at that...

util/mime-type.service.ts
```diff
@@ -0,0 +1,54 @@
+import { Injectable } from 'angular2/core';
+
+@Injectable()
+export class MimeTypeService {
```
cavis: General question: what rule-of-thumb should we use to determine when something goes in `/util` vs `/app/shared`?

util/uploader.ts
```diff
@@ -0,0 +1,323 @@
...
+  constructor(
+    public bucketName:string,
+    public signUrl:string,
+    public awsKey:string,
+    public awsUrl:string = 'https://s3.amazonaws.com',
+    public useCloudfront:boolean = false
+  ) {}
+}
+
+export class Uploader {
```
cavis: What is this `/util/uploader.ts` file about?  Seems to duplicate the upload service, but I don't see where this is used.

kookster: that is not being used at the moment - that is my own ts version of what evaporate does that I was about 75% complete on when I finally got evaporate working. I checked it in here to keep track of if/when I do go back to it.

config/systemjs.config.js
cavis: Hot-reload still works for me.  But I do get a strange `Zone already exported on window the object!` warning from the vendor bundle on each request.

cavis: Also, on a fresh clone-and-npm-install, I'm still getting `systemjs-json` added all over the place.

util/mime-type.service.ts
kookster: idk - we both added them at the same time, so they may just be the same thing, or close enough.
My thought had been that `app/shared` was app specific, where `util/` was stuff that was more generic and not app related. Thinking of it that way, env.ts probably should not be in there!

config/systemjs.config.js
kookster: I'm not worrying about the systemjs-json, seems to be a non-issue, but I am going to figure out this zone downgrade.

app/upload/upload.component.ts
kookster: MimeTypeService is needed here so it can be listed in the providers so sub components can have it injected. HTTP_PROVIDERS is left over from when I was implementing a version of this using the Http service - I'll remove it.

app/upload/upload.service.ts
kookster: moved it to the `Upload` constructor, I'll delete this line.

config/systemjs.config.js
kookster: I've fixed this by upgrading to the latest beta (13), and on jspm upgrade this resolved.

kookster:
I think I've fixed all the issues @cavis brought up (thanks for a good review!)

cavis:
:shipit:

PR #6, opened 2016-03-17
Login and auth

cavis:
Wiring up to prx auth service.

NOTE: to make this work, you have to `echo 3000 > ~/.pow/publish.prx`.  And then after `npm start`-ing, change the URL to http://publish.prx.dev.  Otherwise the oauth-callback won't work.

Included:

- Forces a "Login" page any time you're unauthorized
- Displays any iframe error from id.prx.org as "Invalid username or password" (which is usually true)
- Does not allow logout, yet
- Displays username/avatar in the menu bar
- Basic setup for a CMS-API service.  (I left a bunch of TODO errors, in places I'd rather TDD later on instead of hacking it out now).

Problems:

- Does not play well with browsersync (there's a socket error on every refresh, even though things seem to refresh normally)
- Bad logins also log javascript errors (iframe issues)


src/app/auth/navlogin.component.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  private authUrl: string;
+  private iframe: Element;
+  private isLoading: boolean = true;
+  private authUser: PrxAuthUser;
+
+  constructor(private element: ElementRef, private authService: AuthService) {
+    this.authService.user.subscribe(this.authChanged);
+
+    var host = 'https://id.prx.org';
```
kookster: yeah, looks like candidates for the env config stuff

src/app/login/login.component.css
```diff
@@ -0,0 +1,30 @@
+/*
+ * Login page (mostly an iframe)
+ */
+/*:host {
```
kookster: commented out?

cavis: whoops!  will delete

gulpfile.js
```diff
@@ -38,13 +40,27 @@ gulp.task('git:hooks:install', shell.task([
...
+  return {};
+}
+function encodeDotValue(val) {
+  if (['true', 'false', 'null', 'undefined'].indexOf(val) > -1) return val;
+  if (isNaN(val) || val == '') return "'" + val + "'";
+  return val;
+}
 gulp.task('env:write', function(cb) {
-  var env = dotenv.parse(fs.readFileSync('.env', { encoding: 'utf8' }));
+  var defaults = parseDotFile('env-defaults', true);
```
kookster: Not sure how I feel about this, since it seems it might be a surprise to folks that something other than the `.env*` file is being loaded and used.  I'm gonna leave it be, since I am imagining possible confusion rather than having any factual issue with this, but it is not how most .env stuff gets loaded in other standard libs and other projects.

src/app/shared/auth/auth.component.ts
```diff
@@ -0,0 +1,48 @@
...
+})
+
+export class AuthComponent {
+
+  private authUrl: string;
+
+  // assume user is logged in, to start
+  private isAuthorized: boolean = true;
+
+  constructor(private element: ElementRef, private authService: AuthService, private cmsService: CmsService) {
```
kookster: Seems a bit long, I would probably stack these args

src/app/login/login.component.ts
```diff
@@ -0,0 +1,50 @@
...
+    `
+})
+
+export class LoginComponent {
+
+  private iframeUrl: string;
+  private errorMsg: string;
+  private firstLoad: boolean = false;
+
+  constructor(private element: ElementRef, private authService: AuthService, private cmsService: CmsService) {
```
kookster: stack these?

src/app/shared/cms/cms.service.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.http.get(`${Env.CMS_HOST}/api/v1`).subscribe((res) => {
+      this.rootDoc.next(res.json());
+    });
+  }
+
+  set token(token: string) {
+    this.authToken.next(token);
+  }
+
+  get root(): Observable<HalDoc> {
```
kookster: nota crisis, but it seems like having a root doc is a standard hal thing; I wonder if some of this should be pushed down to `haldoc.ts`?

kookster: could be premature. Just thinking other hal services will end up with similar methods.

src/app/shared/cms/haldoc.ts
```diff
@@ -0,0 +1,97 @@
+import {Http, Headers, RequestOptions, Response} from 'angular2/http';
+import {Observable} from 'rxjs/Observable';
+
+/*
+ * Generic class for interacting with HAL api
+ */
+export class HalDoc {
+  private _http: Http;
```
kookster: I dig the `_privatevar` naming, but it's inconsistent. I think since ts will check this for us, it might be ok to leave it off, which could also make the constructor simpler if the args declared with private properties.

src/app/shared/cms/haldoc.ts
```diff
@@ -0,0 +1,97 @@
...
+        return doc.follows.apply(doc, rels);
+      });
+    }
+    else {
+      return Observable.of(this);
+    }
+  }
+
+  links(rel: string): any[] {
+    if (this['_links'] && this['_links'][rel]) {
```
kookster: This is the other thing, the `_jsonelement` looks like private vars, which is confusing.

kookster: I am easily confused really

src/app/shared/cms/cms.service.ts
cavis: Yeah, i'd prefer to see how things shake out.  My intention here was to provide a single memoized root doc.  But longer term, we'll probably need a full-blown caching strategy somewhere.

src/app/shared/cms/haldoc.ts
cavis: Ah, right.  I factored out a bunch of others, but must have missed that one.

PR #9, opened 2016-03-23
Upgrade to Angular2 beta 11

kookster:


kookster:
ok @cavis , ready now - also gets rid of the deprecation warning I used to get on running tests!

PR #10, opened 2016-03-23
Fix angular2 dependency loading

kookster:


kookster:
this should work, @farski wanna check it out for embed?

farski:
@kookster seeing a few problems trying to do this in backsaw, but idk what else may not be in sync. 

Getting `Uncaught TypeError: Cannot read property 'cancelTask' of undefined` errors on the page, and Atom/TS complains every time I try to load an ng or rx module

kookster:
I saw that error too on publish, but it didn't seem to be causing issues, so wasn't sure what to make of it.

kookster:
@cavis @farski I think this is good to go now.
We still need angular2 and rxjs in the node_modules dir, but the npm tasks now will make sure these are in sync by copying from jspm -> node rather than having to do so ourselves with installs and upgrades

kookster:
@cavis I also added some missing type casts in services, were you not seeing those as errors in atom-typescript?

cavis:
Atom-typescript has been real backed up for me lately - may have just missed it.  Will take a look!

gulpfile.js
```diff
@@ -39,7 +41,7 @@ gulp.task('jspm:unbundle',   shell.task('jspm unbundle'));
 
 // Compile tasks
 gulp.task('jade:index:dev', () => {
-  return gulp
+  return gulp.forEach
```
cavis: What's this `forEach` about?  It's undefined for me, after a fresh npm install.

gulpfile.js
```diff
@@ -1,17 +1,19 @@
...
 const dotenv = require('dotenv');
 const fs     = require('fs');
 const KarmaServer  = require('karma').Server
 
 // Public tasks (serial)
 gulp.task('git:hooks:pre-commit', seq('jspm:unbundle'));
-gulp.task('postinstall',          seq('jspm:install', 'typings:install', 'git:hooks:install'));
-gulp.task('start',                seq('build:dev', 'env:watch', 'server:dev'));
+gulp.task('postinstall',          seq('jspm:install', 'typings:install', 'git:hooks:install', 'copy:deps'));
+gulp.task('start',                seq('build:dev', 'env:watch', 'copy:deps', 'server:dev'));
```
cavis: Do these actually need to be copied when you're starting a dev server?  Or is this just a convenient place to sync them?

cavis:
It works!  (Other than the browsersync socket error)... huzzah!

config/systemjs.config.js
```diff
@@ -32,12 +32,13 @@ System.config({
   },
 
   map: {
-    "angular2": "npm:angular2@2.0.0-beta.11",
+    "angular2": "npm:angular2@2.0.0-beta.12",
```
cavis: So if you upgrade this to `beta.12`, and the package.json version is still at `beta.9`... which one does it pick?

farski: package.json should have ^2.0.0-beta.9, which beta.12 satisfies

gulpfile.js
kookster: that looks like a mistake - will fix.

gulpfile.js
kookster: just convenient so it happens as part of the regular workflow.
open to suggestions.

cavis:
Alright, I think we're good on this then.  :+1:

PR #11, opened 2016-03-24
Hot reloader

cavis:
Hot reload modules


config/systemjs.config.js
```diff
@@ -90,13 +122,18 @@ System.config({
       "fs": "github:jspm/nodelibs-fs@0.1.2",
       "inherits": "npm:inherits@2.0.1",
       "minimalistic-assert": "npm:minimalistic-assert@1.0.0",
+      "systemjs-json": "github:systemjs/plugin-json@0.1.0",
```
kookster: When I did a clean install of npm and jspm modules, I don't get any sign of `systemjs-json` in my `systemjs.config.js`.

@cavis Can you do a clean install? When you do, is this `systemjs-json` still in here?

config/karma.conf.js
```diff
@@ -16,17 +16,20 @@ module.exports = function(config) {
 
     jspm: {
       loadFiles: [
-        'src/**/*.spec.ts'
```
kookster: I went ahead and made these small changes to get tests working again.

server.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+const httpServer = require('http-server');
+const open = require('open');
+
+let cache = 3600;
+if (process.env.NODE_ENV === 'production') {
```
kookster: Is the thought that we would run this node server in production too?

kookster:
Other than the discrepancy between your and my `systemjs.config.js`, this all looks good to me.

kookster:
Here's what my systemjs.config.js looks like:
https://gist.github.com/kookster/1936dfd343969d8a56dc

server.js
cavis: That's the idea.  At some point, I'll probably modify it to have the same "dev vs dist" flag as `index.jade`.

config/systemjs.config.js
cavis: Doing a clean install, the only differences in mine are:

1. Upgrades rxjs to `beta.3` (which we don't want)
2. Removes the top-level zone.js dependency (but leaves the nested ones)

PR #13, opened 2016-03-29
Switch to express server

cavis:
Turns out, there's no way to set a default file (or proxy) in `http-server`.  So you'll never be able to reload a non-root angular route.

This switches to express, and also dynamically compiles `index.html` and `env.ts`, so gulp doesn't have to.


cavis:
Hrmm... actually, that dynamic `env.ts` may not work.

cavis:
Okay, for now, this creates an empty `util/env.ts` for the duration of jspm-bundling.

PR #14, opened 2016-04-01
Story edit pages

cavis:
Puts together the basic structure for story editing (and sub-pages of that).  Also includes a lot of work around CMS integration and testing.


app/app.component.ts
```diff
@@ -8,7 +8,8 @@ import {NavUserComponent} from './header/navuser.component';
 import {FooterComponent}  from './footer/footer.component';
 import {LoginComponent} from './login/login.component';
 
-import {HomeComponent} from './home/home.component';
+import {HomeComponent}     from './home/home.component';
```
kookster: whitespace (I am better than a linter)

app/header/navuser.component.ts
```diff
@@ -33,14 +33,17 @@ export class NavUserComponent {
   private userName: string;
   private userImage: string;
 
-  constructor(private cmsService: CmsService) {
-    this.cmsService.follows('prx:authorization', 'prx:default-account').subscribe((doc) => {
+  constructor(private cms: CmsService) {
+    cms.follow('prx:authorization').follow('prx:default-account').subscribe((doc) => {
```
kookster: I do like this syntax better :+1:

app/home/home.component.ts
```diff
@@ -1,30 +1,48 @@
...
-      <p *ngIf="!userName">You are not logged in. Shame on you.</p>
+    <div class="main">
+      <section>
+        <spinner *ngIf="!accounts"></spinner>
+        <div *ngIf="accounts">
+          <h1>Your Accounts</h1>
+          <div *ngFor="#account of accounts; #i = index">
+            <h2>{{account.name}} <i>{{account.type}}</i></h2>
+            <ul>
+              <li *ngFor="#story of accountStories[i]">
```
kookster: How is this limited - one page, whatever is embedded?

app/shared/cms/haldoc.ts
```diff
@@ -1,98 +1,165 @@
...
-import TemplateParser from 'url-template';
+import {HalRemote} from './halremote';
+
+// Bring observables up to snuff
+export interface HalObservable<T> extends Observable<T> {
+  follow(rel: string, params?: {}): HalObservable<T>;
+  followList(rel: string, params?: {}): HalObservable<T[]>;
+  followItems(rel: string, params?: {}): HalObservable<T[]>;
+}
+Observable.prototype['follow'] = function(rel: string, params: {} = null) {
```
kookster: You're adding these follow* methods to Observable itself? Is this b/c you can't subtype it or something?

app/home/home.component.ts
cavis: It ends up making a GET request to `/authorization/account/:id/stories` for each account.  Which I think just comes back with whatever the CMS kaminari default is?  (10?)

This code in particular is just a "proof of concept" - I'm guessing whomever <cough cough> comes up with the actual UI will do a better job with requests/paging.

app/shared/cms/haldoc.ts
cavis: Yeah, it turned out that extending Observable was really tough in typescript.  (See [this stackoverflow](http://stackoverflow.com/questions/30494708/how-can-i-create-a-class-that-extends-rx-observable-in-typescript)).

But adding to the Observable prototype is what they actually recommend in the rxjs docs.  So I opted to do that, and then use that `HalObservable` interface, which "knows about" the new methods.

app/shared/cms/haldoc.ts
```diff
@@ -1,98 +1,165 @@
...
 
-  followLink(linkObj: any, params: any = {}): Observable<HalDoc> {
-    let headers = new Headers();
-    headers.append('Accept', 'application/json');
-    if (this.token) {
-      headers.append('Authorization', `Bearer ${this.token}`);
+  update(data: any): HalObservable<HalDoc> {
+    let link = this['_links'] ? this['_links']['self'] : null;
+    if (!link) {
+      return <HalObservable<HalDoc>> this.error(`Expected update link at _links.self - got null`);
```
cavis: Here's a consequence of using `HalObservable` - things like `Observable.empty()` and `Observable.error()` also extend Observable.  So you can't directly return an `Observable.error()` without typescript complaining that it can't be cast to a `HalObservable`.

My hack was to return the error from a different method and then cast it.  Makes typescript happy.

app/shared/cms/halremote.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (cachedResponse) {
+        return cachedResponse;
+      } else {
+        return HalRemoteCache.set(href, this.http.get(href, this.httpOptions()).flatMap((res) => {
+          if (res.status === 200) {
+            return Observable.of(res.json());
+          } else {
+            return Observable.throw(new Error(`Got ${res.status} from GET ${href}`));
+          }
+        }));
```
kookster: seems like this pattern of getting from cache, or retrieving and setting in cache, could be implemented in a cache method

kookster: But then again, looks like you aren't doing this often, so perhaps fine to just be in get, and keep the cache simple.

app/shared/cms/halremote.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private httpOptions(hasContent = false): RequestOptionsArgs {
+    let headers = new Headers();
+    headers.append('Accept', 'application/json');
+    if (this.token) {
+      headers.append('Authorization', `Bearer ${this.token}`);
+    }
+    if (hasContent) {
+      headers.append('Content-Type', 'application/json');
```
kookster: any reason to use the hal json type?

app/storyedit/models/story.model.ts
```diff
@@ -0,0 +1,125 @@
...
+  title: boolean = false;
+  shortDescription: boolean = false;
+}
+
+class StoryInvalid {
+  title: string;
+  shortDescription: string;
+}
+
+export class StoryModel {
```
kookster: Duplication in this class for the Payload, Changed, Invalid, and the `set` methods.
Seems like there should be some way to do this with less duplicate code per attribute?

kookster: I also recognize that may be a goal for a later PR.

app/shared/cms/halremote.ts
cavis: This sort of turned into a glorified GET-request-consolidator.  I'd definitely be in favor of refactoring it down the road.

index.jade
```diff
@@ -19,8 +19,12 @@ html(lang='en')
 
             // reload same-named components when css changes
             hr.on('change', (fileName) => {
+              if (console['clear']) console['clear']();
+              console.log(`---- hot-reloaded ${fileName} ----`);
               if (fileName.match(/app\.component/)) {
                 hr.hotReload('app/app.component.ts');
+              } else if (fileName.match(/storyedit\.forms/)) {
+                hr.hotReload('app/storyedit/directives/edit.component.ts');
```
kookster: educate me - why do some ts files need to be added here but not others/most?

app/shared/cms/halremote.ts
cavis: Would that be `application/hal+json`?  Yeah, that would probably be more correct.

app/storyedit/models/story.model.ts
cavis: Yeah, I'm actually fairly unhappy with this model class.  I went through way too much effort to run validations "on field set" rather than "on demand".  Currently refactoring it as I work on putting in tags, into a more normal `isValid('title')` sort of thing.

index.jade
cavis: I'm trying to "guess" which ts file to load, based on the name of the css file that changed.  In many cases, they have the same name except for the `css/ts` extension.  But this form css is shared between components, so for faster debugging I manually triggered the refresh.

In other words... that's mostly a debugging-helper.

kookster:
lgtm 👍

PR #16, opened 2016-04-19
Feat/tagging

cavis:
Init tagging on the first story-edit page.  The designs show this including:

- "Genre" (I'm using the [itunes podcast categories](https://help.apple.com/itc/podcasts_connect/#/itc9267a2f12))
- "Type" (TODO because I'm not sure what this is)
- "Free Tags"

Also added "SubGenre", for itunes subcategories.  All are being stored on the story `tags` array - I just hardcode the ones in the dropdowns.

This PR also includes some refactoring of form components.

### DIFF
--- a/app/storyedit/directives/edit.component.css
+++ b/app/storyedit/directives/edit.component.css
@@ -1,0 +1,22 @@
+/**
+ * Story basic editing
+ */
+hr {
+  width: 100%;
+  border-top: 1px solid #aaa;
+  border-bottom: 1px solid #ccc;
+  margin: 0 0 30px;
+}
+.uploads-box {
+  border: 4px solid purple;
+  height: 300px;
+  text-align: center;
+  line-height: 300px;
+  color: purple;
+}
+@media screen and (min-width: 768px) {
+  .span-fields > * {
+    display: inline-block;
+    width: 40%;
+  }
+}
--- a/app/storyedit/directives/edit.component.ts
+++ b/app/storyedit/directives/edit.component.ts
@@ -1,66 +1,44 @@
 import {Component, Input} from 'angular2/core';
-import {NgForm} from 'angular2/common';
 import {StoryModel} from '../models/story.model';
 import {CATEGORIES, SUBCATEGORIES} from '../models/story.categories';
+import {StoryFieldComponent} from './storyfield.component';
 
 @Component({
-  directives: [NgForm],
+  directives: [StoryFieldComponent],
   selector: 'newstory-edit',
-  styleUrls: ['app/storyedit/storyedit.forms.css'],
+  styleUrls: ['app/storyedit/directives/edit.component.css'],
   template: `
     <form *ngIf="story && story.isLoaded">
 
-      <div [class]="fieldClasses('title')">
-        <h3><label for="title" required>Story Title</label></h3>
-        <p class="hint">Write a short, Tweetable title like a newspaper headline.
-          Make viewers want to click on your piece from the title alone.</p>
-        <textarea id="title" [(ngModel)]="story.title"></textarea>
-        <p class="error">Title {{story.invalid('title')}}</p>
-      </div>
+      <story-field [story]="story" textarea="true" name="title" label="Story Title" required>
+        <hint>Write a short, Tweetable title like a newspaper headline. Make viewers
+          want to click on your piece from the title alone.</hint>
+      </story-field>
 
-      <div [class]="fieldClasses('shortDescription')">
-        <h3><label for="teaser" required>Teaser</label></h3>
-        <p class="hint">A first impression; think of this as the single-item
-          lead of a piece.</p>
-        <textarea id="teaser" [(ngModel)]="story.shortDescription"></textarea>
-        <p class="error">Teaser {{story.invalid('shortDescription')}}</p>
-      </div>
+      <story-field [story]="story" textarea="true" name="shortDescription" label="Teaser" required>
+        <hint>A first impression; think of this as the single-item lead of a piece.</hint>
+      </story-field>
 
       <hr/>
 
-      <div class="field">
-        <h3><label required>Story Versions</label></h3>
+      <story-field label="Story Versions" required>
         <div class="uploads-box">uploads goes heres</div>
-      </div>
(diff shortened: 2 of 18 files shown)


cavis:
Here's what this looks like:

image

There's a shared validation-message.

cavis:
Note: this still isn't working with prod cms caching.  See PRX/cms.prx.org#135.

app/storyedit/directives/edit.component.ts
```diff
@@ -1,62 +1,59 @@
...
+      </story-field>
 
       <hr/>
 
-      <div class="field">
-        <h3><label required>Tag your Story</label></h3>
-        <p class="hint">Adding tags that are relevant to your piece helps people
-          find your work on PRX and can help you be licensed and heard.</p>
-      </div>
+      <story-field [story]="story" label="Tag your Story" invalid="tags" invalidlabel="" required>
```
cavis: This is a bit wonky, but I wanted this group-of-fields to have shared validation messages.  Thus the invalid `@Input`, which ties this to `story.invalid('tags')`.  And the blank `invalidLabel` prevents the story-field from prefixing the validation message with `"Tag your story ${message}"`.

app/storyedit/directives/edit.component.css
```diff
@@ -0,0 +1,22 @@
...
+ * Story basic editing
+ */
+hr {
+  width: 100%;
+  border-top: 1px solid #aaa;
+  border-bottom: 1px solid #ccc;
+  margin: 0 0 30px;
+}
+.uploads-box {
+  border: 4px solid purple;
```
kookster: purple?

app/storyedit/directives/edit.component.ts
kookster: makes sense, the only other way I guess would be a distinct directive/component just for this, and that seems like overkill

app/storyedit/directives/edit.component.ts
```diff
@@ -1,62 +1,59 @@
...
+          on PRX and can help you be licensed and heard.</hint>
+        <div class="span-fields">
+          <story-field [story]="story" [select]="GENRES" name="genre" label="Genre"
+            small="true" required></story-field>
+          <story-field [story]="story" [select]="SUBGENRES" name="subGenre" label="SubGenre"
+            small="true" required></story-field>
+        </div>
+        <story-field [story]="story" textinput="true" name="extraTags"
+          label="Extra Tags" small="true"></story-field>
+      </story-field>
```
kookster: This is just an opinion, but I am unsure how much tagging like this is helpful, since the only thing itunes and others care about are the tags at the podcast/channel level, not the individual story/item level.

The main time these are useful is when a story isn't in a series/podcast, so I am not gonna sweat them too much at the moment.

app/storyedit/directives/hero.component.ts
```diff
@@ -18,18 +20,18 @@ import {SpinnerComponent}  from '../../shared/spinner/spinner.component';
...
           <button class="preview">Preview</button>
-          <button *ngIf="!story.id" class="create"
-            [disabled]="!story.isValid"
+          <button *ngIf="story.isNew" class="create"
+            [disabled]="story.invalid()"
             (click)="save()">Create</button>
-          <button *ngIf="story.id" class="save"
-            [disabled]="!story.isValid || !story.isChanged"
+          <button *ngIf="!story.isNew" class="save"
+            [disabled]="story.invalid() || !story.changed()"
```
kookster: nice. very readable, logical.

app/storyedit/directives/edit.component.ts
cavis: I'm assuming as i add more fields, that might make more sense.  Edit pages 2/3 have more of these shared components.

I'll fix it later, dangit!

app/storyedit/directives/edit.component.css
cavis: don't question my design choices.

app/storyedit/directives/edit.component.ts
cavis: Hmm... interesting.  Well, this component could certainly be copied/pasted to there.

Should I swap these `<option>`s out for something else then?  Or just wait until we know more specifically what we want?

app/storyedit/directives/hero.component.ts
cavis: Yeah, these model-style bindings seem to work pretty well.  Couple of UX issues to address down the road (how/when to show changed/invalid states).  But the bindings are all there if we need them.

app/storyedit/directives/edit.component.ts
kookster: I think these are good for now, and we'll see what we want later.

app/storyedit/models/story.model.ts
```diff
@@ -69,57 +125,59 @@ export class StoryModel {
...
+        return `SubGenre ${this.invalid('subGenre')}`;
+      } else if (this.invalid('extraTags')) {
+        return `Tags ${this.invalid('extraTags')}`;
+      }
+    } else if (!fieldName) {
+      return this.invalid('title') || this.invalid('shortDescription') ||
+             this.invalid('genre') || this.invalid('subGenre');
+    } else {
+      return null;
     }
```
kookster: This seems like a pretty different approach from the ng2 control and control groups stuff, like this or something -  http://blog.thoughtram.io/angular/2016/03/14/custom-validators-in-angular-2.html

Did you try this style of form builder and it was lacking?

cavis: I started out with that template-driven approach.  It supports "most" of the validations I'd need out-of-the-box.  My hangup was that with the state living in the form, it (1) didn't persist across the 3 subpages, and (2) required a bunch of craziness to pass the valid/changed state up to the parent components "Save/Publish" buttons.

There was a middle-ground option to use the `FormBuilder` api, but I found the syntax a bit verbose.  Encapsulating the state in the model ended up being the cleanest way to share bindings on it.

That said, any simpler form would probably be easier to do with the built-in stuff.  Also, this could probably use (another) refactor to get the validations out of the model, into their own `story.validations.ts` module.

PR #18, opened 2016-05-11
Dockerize development

kookster:


Dockerfile
```diff
@@ -26,3 +23,6 @@ RUN apk --update add --virtual build-dependencies git && \
     find /usr/lib/node_modules/npm -name test -o -name .bin -type d | xargs rm -rf ; \
     rm -rf /usr/share/man /tmp/* /root/.npm /root/.node-gyp \
     /usr/lib/node_modules/npm/man /usr/lib/node_modules/npm/doc /usr/lib/node_modules/npm/html)
+
+ENTRYPOINT ["/tini", "--", "./bin/application"]
+CMD [ "server" ]
```
cavis: Does this need to be after the `RUN` because a different command will trigger a rebuild?  Or just for aesthetics?

README.md
```diff
@@ -1,3 +1,60 @@
...
+Over time additional distribution channels are planned such as twitter cards and facebook sharing, CMS plugins for producer/station sites, publishing platforms like medium, tumbler, and wordpress.com, APIs like the PMP (Public Media Platform), youtube, and soundcloud.
+
+The current broadcast distribution will also be merged to allow for licensing, subscribing, and distribution to station automation systems. We are also looking into supporting instant articles and AMP for stories published on PRX.
+
+Once published, the other side of the coin is to collect metrics on these channels, and consolidate to provide a complete picture of a story's impact and audience.
+
+# Install
+
+This local dev installation expects that both cms and id are installed locally.
+You will need to have an client application setup for publish in id.prx.dev
```
cavis: Well, technically you can also leave the defaults and it'll work against prod `id.prx.org` - assuming you're still using `publish.prx.dev`.  Maybe we should update that default to be the production `publish.prx.docker` application?

lib/server-dev.js
```diff
@@ -10,15 +10,14 @@ var config  = require('./config');
  */
 var server = listen(3000);
 emitter({app: server});
-open('http://publish.prx.dev');
```
cavis: I've started to find this annoying as well.  Stop opening that page over and over!

Dockerfile
kookster: It is mostly aesthetics - most of our other docker files (and ones I have read) have this at the end, as it is sort of logically how the image will execute, where above it is more about the actual build.

The way we currently build, while optimal for image size, also always busts the cache, so it doesn't really affect that.

lib/server-dev.js
kookster: doesn't work from the container anyway ;)

bin/application
```diff
@@ -0,0 +1,32 @@
...
+  shift
+  CMD_ARGS=$*
+}
+
+ApplicationRun () {
+  CMD=
+  if [ "$PROCESS" = "help" ] ; then
+    ApplicationUsage
+    exit
+  elif [ "$PROCESS" = "--" ] ; then
```
kookster: This is the only slightly useful part of this, besides some consistency with our other containers.

README.md
kookster: that is true, we should have instructions for both - using local .dev or .org

kookster: actually, the `env-default` is set to use `id.prx.dev`, so I think this is right.

lib/server-dev.js
cavis: worth removing `open` from the package.json entirely?

README.md
cavis: but the id in `env.ts` is setup to use `id.prx.org` with `publish.prx.dev`.

kookster: ah, I must have missed that. Yeah, these should be consistent.
Ok, I'll make `env-example` match `env.ts` defaults, and update the client to be `publish.prx.docker`

lib/server-dev.js
kookster: I'll do it!

cavis:
👍

PR #19, opened 2016-05-13
Atomic upload saving/editing

cavis:
Goals:

- Story edits are immediately saved to localstorage
- Nothing is persisted remotely (story, versions or files) until you hit save
- Ability to upload/edit an entire story before persisting anything

Technical stuff:

- Created a `BaseModel` class, which serves as the UX/business layer on top of a remote `HalDoc`
  - Handles validation, tracking changes, localstorage, cascading to child `HalDoc`, etc.  (For instance, changes/validations/saving on a Story also reflects the AudioVersion and AudioFile docs).
- Uploads continue when moving across pages
- May eventually need to refactor `BaseModel` slightly, to allow getting the has-local-changes status of a story, without having to init the whole `StoryModel`.  Maybe a static method?  `StoryModel.hasLocalChanges(storyId)`


cavis:
Okay, at long last, this is ready to go.  It could definitely use some human testing and UX feedback at some point.  But I think the basics are all here, to support some pretty fine-grained "what's invalid, what's changed" controls.  Plus the local-storing stuff, of course.

app/shared/model/base.model.ts
```diff
@@ -0,0 +1,210 @@
...
+    for (let validator of validators) {
+      let invalidMsg = validator(field, value);
+      if (invalidMsg) {
+        return invalidMsg;
+      }
+    }
+    return null;
+  }
+
+  changidate(field: string, value: any): boolean {
```
kookster: wat? checking if this will be a change?

cavis: Checking if a field has changed from the original (remotely-stored) value.

Naming is hard, okay?  But yes, something like `changecheck` would probably be better than my sarcastic naming.

kookster:
lgtm

PR #21, opened 2016-06-03
Update to beta.17

cavis:
Mostly fixing deprecated stuff, and more major changes in how Router and ViewChild work.

Also got rid of the `jspm:unbundle` pre-commit hook.  Development mode now just creates a 2nd (ignored) `config/systemjs.dev.js` file, which contains just the vendor bundle.


app/footer/footer.component.ts
```diff
@@ -52,12 +51,12 @@ import {Location} from 'angular2/router';
 
 export class FooterComponent {
 
-  constructor(
-    private location: Location
-  ) {}
+  locationPath(): string {
```
farski: this could be a `get`

cavis: For some reason, jasmine's `spyOn` doesn't work on getters.  So if I want to mock it, it has to be an actual function.

This was more straight forward before they took `Location` out of the router.

farski: huh. bummer

farski:
Seems okay to me, will let @kookster review and merge

kookster:
lgtn 👍

PR #22, opened 2016-06-23
Upgrade to RC3

cavis:
Notes:

- Upgrade to rc3
- Changed to the new router
- Refactored my "authorized/unauthorized" guards to work with the new router
- A bit of hacking in the tests, as there aren't test components for the new router yet


farski:
Gave this a once over, but need to come back and dig into a few parts

PR #23, opened 2016-07-14
Dashboard and images

cavis:
- Starts implementing a series-oriented dashboard
- Implement image-uploading for stories
- Lot of bugfixes for story editor

Requires PRX/cms.prx.org#143


app/home/directives/home-series.component.ts
```diff
@@ -0,0 +1,106 @@
...
+import {TimeAgoPipe} from '../../shared/date/timeago.pipe';
+import {HomeStoryComponent} from './home-story.component';
+
+@Component({
+  directives: [SpinnerComponent, ImageLoaderComponent, HomeStoryComponent],
+  pipes: [TimeAgoPipe],
+  selector: 'home-series',
+  styleUrls: ['app/home/directives/home-series.component.css'],
+  template: `
+    <header *ngIf="noseries">
```
kookster: shouldn't we show some standalone stories even if there are series? just lower down the page maybe?

app/home/directives/home-series.component.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  loadSeriesStories() {
+    this.title = this.series['title'];
+    this.count = this.series.count('prx:stories');
+    this.updated = new Date(this.series['updatedAt']);
+    this.series.follow('prx:image').subscribe(
+      img => this.logo = img.expand('enclosure'),
+      err => this.logo = null
```
kookster: not in this PR, but would be good to impl some default (perhaps you have and I haven't noticed it yet)

app/home/directives/home-series.component.ts
cavis: Yep - it does.  This poorly named component displays a list of stories.  The list either has a parent-series or does not (via that `noseries` input).  In the HomeComponent, you'll see a switch that decides how many of these to display (and how many rows of stories) based on if you have a series or not.

app/home/directives/home-series.component.ts
cavis: Agreed.  And I have not added any default images yet.

app/home/directives/home-series.component.ts
kookster: sorry, that is in the home component, not here.

app/home/directives/home-story.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (this.story.isNew) {
+      this.editLink = ['/create'];
+      if (this.story.parent) {
+        this.editLink.push(this.story.parent.id);
+      }
+    } else {
+      this.editLink = ['/edit', this.story.id];
+    }
+
+    // TODO: draft audios
```
kookster: ?

app/storyedit/storyedit.component.ts
```diff
@@ -62,37 +58,75 @@ export class StoryEditComponent implements OnInit, OnDestroy, CanDeactivate<CanD
...
   }
 
   ngOnDestroy() {
     this.routeSub.unsubscribe();
-    this.tabSub.unsubscribe();
   }
 
-  canDeactivate(next: any, prev: any): Observable<boolean> {
-    if (this.story.changed()) {
+  canDeactivate(next: any, prev: any): boolean | Observable<boolean> {
```
kookster: not familiar with this use of a pipe...

app/home/directives/home-story.component.ts
cavis: Related to the lambda stuff I'm working on.  Some way to access/playback the "temporary" upload mp3s, before the lambda-fixer-thing has a chance to move the files to the permanent location.

app/storyedit/storyedit.component.ts
cavis: Just a way to tell typescript that it'll return "one of those two things".  When you use it with an input param, typescript is also smart enough to recognize any `if (thing instanceof Something)` and cast it accordingly.

Issue #30, opened 2016-08-31
placeholder images


cavis:
Implement fallback images for series, story, and account.  Will need two flavors of them:

1. When image isn't set in the API
2. When image is a 404 (happens a lot using the non-prod S3 bucket)

cavis:
Done by PRX/publish.prx.org#37

PR #37, opened 2016-09-07
Feat/image placeholder

sandikbarr:
Implements the placeholder images for ticket #30. The placeholder images are temporary, to be replaced. I used the same no image and error placeholders for story, series, and account, basically wherever ImageLoaderComponent is used, but now I'm second guessing this decision. Did we want to use different placeholders for each of these?


cavis:
(Re-sync'd the snap-ci PR pipelines from master, to get rid of the old jspm commands)

src/app/shared/image/image-loader.component.ts
```diff
@@ -4,18 +4,22 @@ import { Component, Input, ElementRef } from '@angular/core';
   selector: 'image-loader',
   styleUrls: ['image-loader.component.css'],
   template: `
-    <img *ngIf="src" [src]="src" class="loading" (load)="onLoad()"/>
+    <img *ngIf="src" [src]="src" class="loading" (load)="setBackgroundImage(src)" (error)="setBackgroundImage(placeholderError)"/>
+    <img *ngIf="!src" [src]="placeholderImage" class="loading" (load)="setBackgroundImage(placeholderImage)" (error)="setBackgroundImage(placeholderError)"/>
```
cavis: So thinking about the order of events here ...

(1) Ask CMS for the `prx:image`
(2) Component initialized with `[src]=null`
(3) Image loader sets blank/gray background-image, starts loading the placeholder image
(4) Race condition between placeholder-img loading, and CMS returning the `prx:image`
(5) Changes to `[src]=real-image`
(6) Blank/gray background-image, starts loading the real image

Maybe this is an edge case, since CMS usually bundles the `prx:image` in the same json response as the parent story/series/account.  Or maybe the image-loader needs to be smart enough to know if CMS is still loading vs there really isn't an image.  (Null vs undefined?  Or some other explicit attribute?)

cavis:
Nice use of royalty-free images.

src/app/shared/image/image-loader.component.ts
cavis: Also ... even though I was the one to add that `loading` class, I'm pretty sure it does nothing.

cavis:
Related existing bug: still seeing at least one `Unable to find rel prx:image` in the console, coming from the home-series component.  Should also audit to make sure any `follow('prx:image')` are wrapped in an if `has('prx:image')` for this PR.  (As long as we're fixing image-loading problems).

cavis:
@sandikbarr couple follow-ups above

sandikbarr:
@cavis I refactored to the ImageLoaderComponent to handle the prx:image. I didn't like what I did in [33f3709](https://github.com/PRX/publish.prx.org/pull/37/commits/33f3709037095fa70c7abc8d7dd8283f8bc9f0ce) to handle all the cases in the HomeStoryComponent. Let me know what you think of that.

src/app/header/navuser.component.spec.ts
```diff
@@ -18,19 +19,19 @@ describe('NavUserComponent', () => {
       expect(el.querySelector('spinner')).toBeNull();
       expect(el.querySelector('.name').textContent).toEqual('TheAccountName');
       expect(navuser.userName).toEqual('TheAccountName');
-      expect(navuser.userImage).toEqual('http://cms.mock/v1/theimage');
+      expect(el.querySelector('image-loader > img').src).toEqual('http://cms.mock/v1/theimage');
```
sandikbarr: Not sure this is still valid to test here

cavis: It's hard not to violate "separation of concerns" in these component tests.  But any test is better than nothing!

src/app/shared/image/image-loader.component.ts
```diff
@@ -1,21 +1,42 @@
...
-    this.element.nativeElement.style['background-image'] = `url(${this.src})`;
+  setBackgroundImage(src) {
+    this.element.nativeElement.style['background-image'] = `url(${src})`;
+  }
+
+  onLoad = () => this.setBackgroundImage(this.src);
+
+  onError = () => this.setBackgroundImage(ImageLoaderComponent.PLACEHOLDER_ERROR);
+
+  ngOnChanges(changes:any) {
```
cavis: Haven't used this one before.  But I'm a bit surprised it works without the `class ImageLoaderComponent implements OnChanges` in there.  I guess the angular docs have lied to me again!

src/app/shared/image/image-loader.component.ts
```diff
@@ -1,21 +1,42 @@
...
   template: `
-    <img *ngIf="src" [src]="src" class="loading" (load)="onLoad()"/>
+    <img *ngIf="src" [src]="src" (load)="onLoad()" (error)="onError()"/>
     `
 })
 
 export class ImageLoaderComponent {
 
   @Input() public src: string;
+  @Input() public imageDoc: HalDoc;
```
cavis: Yeah, I like the idea of just giving the component the doc it's waiting for.  And assuming our images stay standardized, it will always be under `prx:image -> enclosure`.

src/app/shared/image/image-loader.component.ts
sandikbarr: Yeah I just don't know better. I'll fix it if that's standard.

cavis: Almost seems like a bug on their side, given [the doc example](https://angular.io/docs/ts/latest/api/core/index/OnChanges-class.html).  Yeah, may be worth adding that just in case they fix it someday.

src/app/shared/image/image-loader.component.ts
cavis: Also, if I'm reading this right, you can still just pass in a `src` right?  It just has to be non-null on init, or you'll get the placeholder image.

sandikbarr: Yep, the HeroComponent is still using just src.

cavis:
Just needs the imports/implements for `OnChanges`.  Otherwise, 👍 .

src/app/shared/image/image-loader.component.ts
```diff
@@ -28,14 +28,18 @@ export class ImageLoaderComponent implements OnChanges {
...
+    if (doesntHaveSrc(changes)) {
+      if (hasPRXImageDoc(changes)) {
+        this.imageDoc.follow('prx:image').subscribe(
+          img => this.src = img.expand('enclosure'),
+          err => this.src = ImageLoaderComponent.PLACEHOLDER_ERROR
+        );
+      } else {
+        this.src = ImageLoaderComponent.PLACEHOLDER_IMAGE;
+      }
     }
```
sandikbarr: This seems to read a bit better :)

sandikbarr:
I lied about the hero image using src. I screwed that up. Fixed.

kookster:
pull requests! ng2 doc mistakes! lies!
excellent. now we're getting something done.

cavis:
Lies and deceit are an important part of pull requests. :shipit:

PR #38, opened 2016-09-12
Series create/edit pages

### LINKED ISSUE #33: series edit page
Ability to create/update a series.

- [x] Edit the standard fields (title, description, etc)
- [x] Upload the series image (limit to a single image)
- [x] Display basic metadata (# of stories, last-updated, etc)

And not required for Guardia epic:

- [ ] Manage the series templates (see PRX/cms.prx.org#147)
  - Audio file and version templates, in the CMS
  - Set basic validations for audio-file length, # of files, etc
- [ ] Manage the series distributions, currently just:
  - PRX.org (required, alway ON)
  - Feeder / "Podcast" (default to ON) (see PRX/feeder.prx.org#129)
  - PRX embed player (default to ON)
- [ ] List the accounts that can edit the series

cavis:
Basic series creation and editing.

Partially implements #33.

- Refactors a bunch of the story-edit-hero code into shared components
- Init a localstorage saveable/discardable Series page, with title/teaser/description fields
- Prevents editing non-v4 stories (though this is commented out until the API starts returning it).


src/app/series/series.component.css
```diff
@@ -0,0 +1,7 @@
+/*
+ * Series edit styling
+ */
+form {
+  padding: 20px 0;
+  max-width: 800px;
+}
```
sandikbarr: Still pretty new to this project, so I'm just asking here. Seeing custom styles for components that seem fairly general on the surface doesn't feel so good. I'm wondering what the eventual roadmap looks like for styling the app. Does it involve an external style guide at all that could be pulled into our various Angular projects?

src/app/series/series.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor(
+    private cms: CmsService,
+    private modal: ModalService,
+    private route: ActivatedRoute,
+    private router: Router
+  ) {}
+
+  ngOnInit() {
+    this.routeSub = this.route.params.subscribe(params => {
+     this.id = +params['id'];
```
sandikbarr: guessing we don't want the + with params['id']

cavis: In this case, I think without that typescript was complaining about me assigning a string (angular parses route params as strings) to a number.

cavis: (I also may have copied that one from the docs -> https://angular.io/docs/ts/latest/guide/router.html#!#get-route-parameter)

sandikbarr: Thanks for the link. This is turning into more of a learning PR for me.

sandikbarr:
Looks good to me. Thanks for having me learn from and poke at all this new-to-me stuff.

Issue #31, opened 2016-08-31
story search page


cavis:
Table/grid view for stories, with a bunch of fancy filters.  Will be linked to from the dashboard ("view all 9 stories").

Display:

- [x] Title
- [x] Short description
- [x] Series title/link
- [x] Image thumbnail (fallback to series)
- [x] Number of audio files, and total duration
- [x] Published/unpublished status
- [ ] ~~Has-unsaved-changes indicator (localstorage)~~
- [x] Last modified date

Filters:

- [x] Always filter to stories editable by the current user
- [x] Simple text filter against story title/description
- [x] Dropdown to filter by series
- [x] Sort by: last-modified, published-date, title

- [x] “View All 99 >” links on the homepage to link into search which will pre-select a series from the dropdown either through more sub-routes (`/search/series/1234`), or some query params (`/search?series=1234`)

TODO: Cards need a bit more design work. The core issue I seem to be struggling with on design is that cards should typically have a single action. Here we've got several actions, play the piece, link to edit the piece, link to edit the series, but it seems like it's not doing a good job of identifying the primary and secondary/tertiary actions.

TODO: debounce search text

PR #44, opened 2016-09-19
story search

sandikbarr:
Adds #31 story search page

TODO: search on search term, ordering

I forgot the orange play button and will add that next.

I could probably use some critical feedback on the card layout. I made them the width that would fit three in a full sized row, which meant I increased per page size to 12. I allowed 2 lines for both story and series titles. If there are tags, the description should be reduced to 2 lines rather than 3. How does all this look with more realistic data?
screen shot 2016-09-21 at 5 18 41 pm


src/app/search/directives/search-story.component.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  ngOnInit() {
+    this.storyId = this.story.id;
+    this.isEmptyDraft = this.story.isNew && !this.story.isStored();
+    this.storyTitle = this.story.title;
+    this.storyUpdated = this.story.lastStored || this.story.updatedAt;
+    this.storyDescription = this.story.shortDescription;
+    this.storyTags = this.story.extraTags && this.story.extraTags.length > 0 ? this.story.extraTags.split(', ') : [];
+
+    if (this.story.isNew) {
```
cavis: I was thinking that "new" stories (only saved in localstorage) wouldn't be included on this page.  Just to avoid the trouble of attempting to injecting them into the persisted CMS results.

So the only place a "new" story would be accessible, would be on the Home page, where they're always listed before the saved stories in the series.

I think.

src/app/search/search.component.html
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+
+      <div>
+        <button class="tab">Series</button>
+        <button (click)="search(1)" class="tab active" disabled>Stories</button>
+
+        <span class="pages">
+          <button *ngIf="totalCount && pages?.length > 1" (click)="search(1, totalCount)" class="btn-link">View All {{totalCount}} &raquo;</button>
+          <button disabled *ngIf="totalCount && pages?.length === 1" class="btn-link">Viewing {{totalCount > 1 ? 'all ' + totalCount : totalCount}}</button>
+          <!-- How is the user supposed to get back to the paged view? -->
```
cavis: Do they need to get back to the paged view?  Altering the search should trigger this button to re-enable, right?

src/app/search/search.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  search(page: number, per: number = 10) {
+    this.currentPage = page;
+    this.isLoaded = false;
+    this.noStories = false;
+
+    let parent = this.searchSeries ? this.searchSeries.doc : this.auth;
+
+    let draftStory = new StoryModel(this.auth);
+    let isUnsavedDraft = draftStory.isNew && draftStory.isStored() && this.searchSeries === null;
+    let showUnsavedDraft = isUnsavedDraft && this.currentPage === 1 && this.searchSeries === null;
```
cavis: Yeah, unlike the homepage, I think you can just exclude drafts here.  Too complicated to apply things like the text-search, series-picker, or sorting to them.

src/app/search/search.component.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (storiesCount > 0) {
+      this.storyLoaders = Array(storiesCount);
+      parent.followItems('prx:stories', params).subscribe((stories) => {
+        this.stories = showUnsavedDraft ? [draftStory] : [];
+        this.storyLoaders = null;
+        let storyDocs = <HalDoc[]> stories;
+        for (let doc of storyDocs) {
+          let story = new StoryModel(parent, doc, false);
+          if (story.doc.has('prx:series')) {
+            // series are embedded, so this should be ok. would prefer to get the series id and use allSeries[id].doc
```
cavis: Yeah, it's a bit wonky.  But I can't think of a better way to do this, with the way the Series model is structured.

src/app/story/directives/hero.component.ts
```diff
@@ -69,7 +69,7 @@ export class StoryHeroComponent implements OnDestroy {
         this.bannerLink = `/series/${this.story.parent.id}`;
         this.bannerLogoDoc = this.story.parent;
       } else {
-        this.bannerTitle = this.story.account['name'] || '(Unnnamed Account)';
```
cavis: Spell check'd.

src/app/search/search.component.html
sandikbarr: yeah, a new search starts on page 1

src/app/search/search.component.ts
sandikbarr: Cool. Drafts do make this more complicated than it needs to be. I was going off the requirement Has-unsaved-changes indicator (localstorage)

src/app/search/directives/search-story.component.ts
sandikbarr: ok. I added them because of this requirement: Has-unsaved-changes indicator (localstorage)

src/app/search/search.component.html
```diff
@@ -0,0 +1,60 @@
...
+        <input (change)="search(1)" [(ngModel)]="searchTerm" placeholder="search by title or description"/><!-- TODO: delay search until finish typing -->
+      </div>
+
+      <div class="form-group">
+        <span class="left">
+          <label [attr.for]="searchSeries">Filter by Series</label>
+          <select id="searchSeries" [(ngModel)]="searchSeriesId" (ngModelChange)="searchBySeries()">
+            <option *ngFor="let seriesId of allSeriesIds" [value]="seriesId">{{allSeries[seriesId] ? allSeries[seriesId].title : 'No Series'}}</option>
+          </select>
+          <!-- TODO: there should be a way to clear this -->
```
sandikbarr: Sigh. I miss react-select. Are there any designs on doing something like this for selects? https://plnkr.co/edit/JcG8uO9nIfSGMEKdLf0Y?p=preview A select component like this would take care of the can't clear the selection problem, and the multi select might be good for tag based searching.

src/app/search/search.component.css
```diff
@@ -0,0 +1,102 @@
...
+}
+
+input {
+  width: 100%;
+  border: 0;
+  padding: 6px;
+  color: #939393;
+  background: #fff;
+}
+
```
sandikbarr: A lot of this stuff could be fairly generic. Should .form-group, .btn-link, and the [disabled] cursor go in forms.css?

cavis: Yeah, anything that may eventually end up in the "style guide" show go into one of those global css files for now.

src/app/search/search.component.html
cavis: If that `angular2-select` works, I'd say go for it.  There are a couple other competing select2 ports ... `ng2-select2`, `angular2-select2`.

The most popular is probably `ng2-select` by valor-software ... but it looks like it may not be compatible with the 2.0.0-final release yet.  (I see an issue for it, though).

sandikbarr: Waiting for something that is more popular and better supported is probably our safest choice.

src/app/search/directives/search-story.component.css
```diff
@@ -0,0 +1,121 @@
...
+  padding: 10px;
+  width: 200px;
+  height: 120px;
+  line-height: 20px;
+}
+.story-title {
+  font-size: 16px;
+  max-height: 40px;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
sandikbarr: This is a lie. Multiline truncation won't ellipsis with text-overflow: ellipsis alone. Do we even want ellipsis in the titles? Two lines seems like enough. I was primarily concerned with overflowing.

src/app/search/directives/search-story.component.css
```diff
@@ -0,0 +1,121 @@
...
+.duration, .modified, .audio-total {
+  width: calc(100%/3);
+}
+.modified {
+  text-align: right;
+}
+.audio-total {
+}
+.audio-total > i {
+  /* TODO: this isn't the correct icon */
```
sandikbarr: TBH, I wasn't even sure if the icon in the mockup was referring to audio file count.

src/app/search/directives/search-story.component.css
```diff
@@ -0,0 +1,121 @@
...
+.story-tags {
+  clear: left;
+  display: flex;
+  justify-content: flex-start;
+  flex-flow: row wrap;
+  width: 320px;
+  padding: 0 10px;
+  max-height: 20px;
+  line-height: 20px;
+  overflow: hidden;
```
sandikbarr: We probably do want ellipsis here though if there are more tags than will fit into the single line that I allowed.

src/app/search/directives/search-story.component.css
```diff
@@ -0,0 +1,121 @@
...
+
+.three-lines {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-height: 60px;
+  line-height: 20px;
+  -webkit-line-clamp: 3;
+  display: block;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
```
sandikbarr: I need to make sure this has been cross browser tested. Do we have specific browser requirements other than all major browsers?

kookster: Major and modern (current and last prior major release), so no old IE support pain.
h/t @farski, I couldn't remember where we wrote this down: https://help.prx.org/hc/en-us/articles/203763430-Supported-browsers

Should probably add this and other FE standards to: 
https://github.com/PRX/meta.prx.org/wiki/Best-Practices for browser testing requirement
and
https://github.com/PRX/meta.prx.org/wiki/Project-Standards needs more FE focused standards like browser support

sandikbarr: This will only truncate with ellipsis in webkit browsers. The fallback is truncation without ellipsis. The mockup actually has a Read More link. I wasn't sure what the intention was there, so I didn't add it yet. Cross browser multiline ellipsis looks kind of gross http://codepen.io/romanrudenko/pen/ymHFh, so I'm gonna leave this alone for now until figuring out what Read More or whatever this truncation should actually do.
screen shot 2016-09-23 at 1 19 25 pm

src/app/search/directives/search-story.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    <div>
+      <section class="story-image">
+        <image-loader [imageDoc]="story.doc"></image-loader>
+        <p *ngIf="statusClass" [class]="statusClass">{{statusText}}</p>
+      </section>
+      <section class="story-detail">
+        <h2 class="story-title"><a [routerLink]="editStoryLink">{{storyTitle}}</a></h2>
+        
+        <h3 *ngIf="seriesLink" class="series-title"><a [routerLink]="seriesLink">{{seriesTitle}}</a></h3>
+        <h3 *ngIf="!seriesLink" class="series-title">{{seriesTitle}}</h3>
```
sandikbarr: This doesn't really stand out as the series title to me. It needs something, maybe an icon.

sandikbarr:
- Angular upgrade is merged into this branch as well now and search is a module like stories and series. Hopefully this will merge cleanly after that one (which I'm not sure if we need to upgrade the angular-cli on staging for first?)
- Should support search text
- Adds orange play icon
- Multiline CSS overflow with ellipsis is dumb/hard

src/app/search/directives/search-story.component.css
cavis: Me either.  This sounds like a job for ... user feedback!

src/app/search/directives/search-story.component.css
cavis: Agreed.  But we can always catch that one later!

And remember for podcasts specifically, story-tags probably aren't as important as series-tags.  (The series will eventually be the source for the itunes tags, when we get around to implementing that).

src/app/search/search.component.html
```diff
@@ -0,0 +1,62 @@
+<div class="main">
+  <section>
+    <header>
+      <h2>Do a search for...</h2>
+
+      <div class="form-group">
+        <input (keyup)="search()" [(ngModel)]="searchText" placeholder="search by title or description"/><!-- TODO: delay search until finish typing -->
```
cavis: yeah, this definitely needs a debounce, at some point.

src/app/search/search.component.ts
```diff
@@ -0,0 +1,142 @@
...
+      filters.push('noseries');
+    }
+    if (this.searchText) {
+      filters.push('text=' + this.searchText);
+    }
+    let params = {page: this.currentPage, per, filters: filters.join(',')};
+
+    let storiesCount = parent.count('prx:stories') || 0; // wrong, doesn't account for filter. looks obvs wrong with No Series filter
+    if (storiesCount > 0) {
+      this.storyLoaders = Array(storiesCount);
```
cavis: Looks like `storyLoaders` never gets reset, if the current search is returning 0 stories.  (So you get > 1 loading-spinners as you run the search).

src/app/search/search.component.html
cavis: In fact, I sometimes get search results back out-of-order when typing.  May even need to cancel any in-flight search requests when you change your search here.  Though the CMS stuff doesn't support that, currently.

Issue #35, opened 2016-09-01
upgrade to angular 2 final


cavis:
- [x] Upgrade dependencies
- [x] Get off the now-entirely-deprecated forms module

cavis:
For dependencies, something like this:

![image](https://cloud.githubusercontent.com/assets/1410587/18283002/9bce92cc-7420-11e6-8adf-374026802f79.png)

Though when `angular-cli` updates, that may be the easier option.

kookster:
merged.

PR #46, opened 2016-09-20
Upgrade to angular final

cavis:
See #35.  Still very much TODO: get the tests working again.

Upgrade to the latest `angular-cli` ... which in turn upgrades us to the final `angular@2.0.0` release.

Notable things addressed:

- Components no longer have `directives` or `pipes`.  These imports must be moved up to the Module level.
- Minor changes to routing and router-directives (mostly organizational).


cavis:
Summary of changes:

- Upgraded `package.json` to new angular-cli, re-ran `ng init`, and resolved conflicts.  (That init command overwrites a bunch of files with the boilerplate again)
- All the "singleton" services/components moved into the `core` module.  This module is only imported once, into the root `app` module, preventing multiple instances of things like the auth/cms services.
- The new trimmed-down `shared` directory is now its own module.
  - NOTE: because of some circular dependency issues between `core` and `shared`, I had to make sure `core` didn't use anything in shared, which resulted in some semi-wonky code in `app.component.html`.  (Specifically, `nav-user.component` can't directly use `image-loader` anymore).
- Story and Series moved into their own modules (and the routes now stay in that module, instead of being imported into the root-routes file).
- Renamed some selectors with the `publish-` prefix to satisfy tslint.

cavis:
Note that if you have angular-cli installed globally, you'll need to:

```
npm uninstall -g angular-cli
npm cache clean
npm install -g angular-cli@1.0.0-beta.15
```

And either way, you should:

```
rm -rf node_modules dist tmp
npm install
```

src/app/series/series.component.html
```diff
@@ -1,14 +1,12 @@
...
+  <h1 class="hero-title" *ngIf="id">Edit Series</h1>
+  <h1 class="hero-title" *ngIf="!id">Create Series</h1>
+  <div class="hero-info" *ngIf="series">
     <h2>{{series.title || '(Untitled)'}}</h2>
     <p *ngIf="series?.isNew">Not saved</p>
     <p *ngIf="!series?.isNew">Last modified {{series.updatedAt | timeago}}</p>
-  </hero-info>
-  <hero-actions *ngIf="series">
+  </div>
+  <div class="hero-actions" *ngIf="series">
```
cavis: They made [a change in RC6](https://github.com/angular/angular/issues/11251) that started throwing warnings for any unknown element names not covered by a component-selector.  So these `ng-content` selectors had to be switched to something else (at least for now - that github issue hints they may have other options down the line).

src/app/shared/date/index.ts
```diff
@@ -0,0 +1 @@
+export * from './timeago.pipe';
```
cavis: Hrmm... yeah, I don't have a good reason.  I've written that `time-ago-in-words` code too many times in too many languages.  Will pull in moment!

cavis:
Shouldn't break staging - the SnapCI docker-build will create a new image with the new version of the CLI on it.

Issue #32, opened 2016-08-31
series search page


cavis:
Table/grid view for series, with a bunch of fancy filters. Will be linked to from the dashboard, in the case that you have more series than the dashboard is willing to display (I think it's 3, right now).

This may be lower priority, since having > 3 podcast-series is an edge case at the moment.

PR #50, opened 2016-09-29
Series Search

sandikbarr:
Implements series search page #32 

Also refactors the original story search work into child components and adds sorting for stories.

Items this does not address:
- debouncing the search text so it waits to search until the user is done typing
- updating url with search or page change
- multiline overflow with ellipsis in non webkit browsers
- the Read More link for truncated descriptions (I don't know what the Read More link should do, and I also have some design concerns about this functionality because not only can the description be truncated but titles and tags too, and we haven't considered a way to expand those)
- unfortunate flex growth of odd items in the last row
- clearing of select dropdowns to remove search parameters
- functional search by genre/subgenre (and where this should actually be applied, stories or series)


src/app/search/search.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  provide(AuthService, {token: authToken, setToken: (t) => cmsToken = t});
+  provide(CmsService, cms);
+  provide(Router, RouterStub);
+  provide(ActivatedRoute, activatedRoute);
+
+  let auth;
+  beforeEach(() => {
+    activatedRoute.testParams = {};
+    auth = cms.mock('prx:authorization', {});
```
sandikbarr: @cavis I'm having a bad time getting past the error 'Un-mocked request for rel prx:authorization'. Any ideas on what I might be missing?

cavis: The testing helpers should provide a lot of those services "by default" for any test starting with `cit`.  So you should be able to remove your `AuthService`, `CmsService`, `Router` and `ActivatedRoute` stubs above.

sandikbarr: I did add the router.stub upon getting the 'No provider for Router!' error, but yeah looks like I can remove AuthService and CmsService

cavis: Ah, yeah, you're right.  I added a global stub for the `routerLink` directives ... but not for the router service itself.

src/app/search/directives/search-series.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  searchGenre: string;
+  searchSubGenre: string;
+
+  searchByText() {
+    this.searchSeriesByText.emit(this.searchText);
+  }
+
+  searchByGenre() {
+    this.searchSeriesByGenre.emit({genre: this.searchGenre, subgenre: this.searchSubGenre});
+  }
```
sandikbarr: I put this search by genre/subgenre in here for series because it's in the mock up and because I thought that @kookster said tags should apply to stories and genre/subgenre to series (for iTunes) but then again I'm not sure I understood that correctly after seeing @cavis https://github.com/PRX/publish.prx.org/pull/44#discussion_r80496575

src/app/search/directives/series-card.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  ngOnInit() {
+    this.seriesId = this.series.id;
+    this.seriesTitle = this.series.title;
+    this.seriesEpisodeCount = this.series.doc.count('prx:stories') || 0;
+
+    this.seriesDescription = this.series.shortDescription;
+    /* TODO: sort out genre/tags, thought Andrew said stories have tags and series have genre
+    this.seriesGenre = this.series.genre;
+    this.seriesSubGenre = this.series.subgenre;*/
```
sandikbarr: Currently displays nothing for genre, see above comment regarding genre vs tags for series vs stories. I can remove this if you think it causes confusion since we aren't supporting it yet.

src/app/search/directives/series-list.component.css
```diff
@@ -0,0 +1,43 @@
...
+  box-flex: 2;
+  -webkit-flex-grow: 2;
+  flex-grow: 2;
+  -webkit-flex-shrink: 1;
+  flex-shrink: 1;
+  height: 192px;
+  background: #fff;
+}
+
+/* attempt to fix the odd last item stretching
```
sandikbarr: I wanted adaptive cards, I got adaptive cards. I don't yet know what to do about the last odd card that grows to the size of the row. It is not a good look.

src/app/search/directives/story-list.component.css
```diff
@@ -0,0 +1,49 @@
...
+  box-flex: 3;
+  -webkit-flex-grow: 3;
+  flex-grow: 3;
+  -webkit-flex-shrink: 1;
+  flex-shrink: 1;
+  height: 200px;
+  background: #fff;
+}
+
+/* not a great attempt to fix the odd last items stretching
```
sandikbarr: similar to last row odd items stretching for series only worse for stories because the base size has 3 items in a row

src/app/search/search.component.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    auth.mock('prx:default-account', {});
+    auth.mockItems('prx:series', []);
+    auth.mockItems('prx:stories', []);
+  });
+
+  cit('detects no results', (fix, el, comp) => {
+    expect(el).not.toContainText('Viewing all');
+    expect(comp.noResults).toBe(true);
+  });
```
sandikbarr: Testing probably needs more work. I found that I couldn't really test what's inside the child components, such as looking for the No stories match your search text. Yet, I'm not sure how much sense it makes sense to test those child components individually because they don't have a lot of responsibility other than displaying what was given to them and emit up events to the parent.

src/app/search/directives/search-series.component.ts
```diff
@@ -0,0 +1,74 @@
+import { Component, Input, Output, EventEmitter } from '@angular/core';
+import { CATEGORIES, SUBCATEGORIES } from '../../shared/model/story.categories';
+
+@Component({
+  selector: 'publish-search-series',
+  styleUrls: ['search-series.component.css'],
+  template: `
+    <div class="form-group">
+      <!-- TODO: delay search until finish typing -->
+      <input (keyup)="searchByText()" [(ngModel)]="searchText" placeholder="search by title or description"/>
```
cavis: There are some interesting ways to debounce this in RXJS.  (Searching "angular2 debounce" gives some results from RC5, which still may be applicable).

cavis: Also - why `keyup` and not `ngModelChange`, as in the rest of your fields?

src/app/search/directives/search-series.component.ts
cavis: Yep, eventually the itunes categories (genre/subgenre here) will be on the series.  It's just not in the data model yet.  But this is a good placeholder/reminder for that.

src/app/search/directives/search-story.component.ts
```diff
@@ -1,96 +1,59 @@
...
+      <input (keyup)="searchByText()" [(ngModel)]="searchText" placeholder="search by title or description"/>
+    </div>
+    
+    <div class="form-group">
+      <p class="left">
+        <label [attr.for]="searchSeries">Filter by Series</label>
+        <select id="searchSeries" [(ngModel)]="searchSeriesId" (ngModelChange)="searchBySeries()">
+          <option *ngFor="let seriesId of allSeriesIds" [value]="seriesId">{{allSeries[seriesId]?.title || 'No Series'}}</option>
+        </select>
+        <!-- TODO: there should be a way to clear this -->
```
cavis: Include a blank/null in `allSeriesIds`?

src/app/search/directives/series-card.component.ts
cavis: TODOs are good.  And I think we'll probably need the itunes-categories / series-genre schema change before this goes live.

src/app/search/directives/series-list.component.css
cavis: Hmm... could you insert a couple of extra 0-height cards at the end of the array?  If one of them gets shoved to the last line, it wouldn't be noticeable.  And otherwise they'd just push the last real-item to the left.

src/app/search/directives/story-card.component.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  statusClass: string;
+  statusText: string;
+
+  ngOnInit() {
+    this.storyId = this.story.id;
+    this.storyTitle = this.story.title;
+    this.storyUpdated = this.story.lastStored || this.story.updatedAt;
+    this.storyDescription = this.story.shortDescription;
+    this.storyTags = this.story.extraTags && this.story.extraTags.length > 0 ? this.story.extraTags.split(', ') : [];
```
cavis: Should there be a space in this `split(', ')` ?

src/app/search/directives/story-card.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.editStoryLink = ['/story', this.story.id];
+
+    if (this.story.doc.has('prx:audio')) {
+      this.story.doc.followItems('prx:audio').subscribe((audios) => {
+        let audiosDocs = <HalDoc[]> audios;
+        if (!audios || audios.length < 1) {
+          this.storyDuration = 0;
+        } else {
+          this.storyAudioTotal = audiosDocs[0].total();
+          this.storyDuration = audios.map((audio) => {
```
cavis: We should probably DRY this stuff up at some point.  But it's probably better to do that when cleaning up the story-edit pages, rather than now.

src/app/search/search.component.ts
```diff
@@ -11,132 +11,265 @@ import { StoryModel, SeriesModel } from '../shared';
...
+
   constructor(private cms: CmsService,
+              private router: Router,
               private route: ActivatedRoute) {}
 
   ngOnInit() {
-    this.route.params.subscribe(params => {
-      this.selectedTab = params['tab'];
-      this.searchSeriesId = +params['seriesId'];
+    this.route.params.forEach((params) => {
```
cavis: Wait, route params are no longer an observable?  When did that happen?  (Going to have to update some of my code, apparently).

src/app/search/search.component.ts
```diff
@@ -11,132 +11,265 @@ import { StoryModel, SeriesModel } from '../shared';
...
-        auth.followItems('prx:series', {filters: 'v4'}).subscribe((series) => {
-          for (let s of <HalDoc[]> series) {
-            this.allSeriesIds.push(s.id);
-            this.allSeries[s.id] = new SeriesModel(auth, s, false);
-          }
+  initStorySearch() {
+    this.allSeriesIds = [-1];
+    this.allSeries = {};
+    this.auth.followItems('prx:series', {filters: 'v4'}).subscribe((series) => {
+      for (let s of <HalDoc[]> series) {
```
cavis: Doesn't really hurt anything ... but I'm not sure you need this cast here.  `followItems` should already be returning a `Haldoc[]`.

src/assets/styles/layout.css
```diff
@@ -22,6 +22,7 @@ article .asides aside > section {
   color: #fff;
 }
 main {
+  display: block;/* IE 11 gives auto/no height to main */
```
cavis: This wouldn't be a problem if you'd stop testing in IE11.  I like to save that for a couple days after the release!

sandikbarr:
@cavis I'll respond to your comments, then let's make a decision about merging or continuing here.

src/app/search/directives/series-list.component.css
sandikbarr: I had a similar thought, but I didn't think of zero height. Seems like an easy fix, let's get it in there.

src/app/search/directives/search-series.component.ts
sandikbarr: Hmm, I think `keyup` is because at first I used the `change` event and it didn't update until the field lost focus. `ngModelChange` seems better than event binding. I made an attempt at debounce like this example https://angular.io/docs/ts/latest/guide/server-communication.html#!#more-observables, but I screwed it up and reverted. It's definitely needed though. Seems like it's gonna take some refactoring, separate PR?

src/app/search/directives/search-story.component.ts
sandikbarr: ...but that feels so 2005 :smiley: How about opening a ticket for adding a select2 implementation in place of selects?

src/app/search/directives/story-card.component.ts
sandikbarr: Yeah, we're adding the space in the StoryModel parseExtraTags

src/app/search/search.component.ts
sandikbarr: The docs have officially changed. Subscribe still works. I changed this when following the testing examples.

sandikbarr:
re: When text searching, the screen briefly flashes with a bunch of extra blank cards, before the result returns and the correct number of cards renders.
This is because the loaders count is based on the count('prx:stories') from its parent, but that count does account for the filters. I first noticed it when I added the v4 filtering. I don't know how to get the right count for loaders based on the filters until I've got the response.

sandikbarr:
re: Filter-by-series doesn't update the location. (Eventually would be nice if every change to the search params was reflected in the ;search params).
I agree that it's needed and I can definitely add this, I just wasn't sure where to stop. Should searchText update the URL too, pages, genre and subgenre? I found that I'm also not saving state when the user has clicked View All records, which is really unfortunate when they change the order by or direction and it goes back to the paged view. Should we open a separate ticket and decide what state goes in the URL?

src/app/search/directives/story-card.component.ts
cavis: Ah, you're right.  i was thinking that was an array.

cavis:
Probably more a UX question ... but when the search changes, I would probably opt to (1) remove all the current cards, and (2) show a single blank-card with a spinner on it.  I see what you mean about not being able to get a count, though.  Maybe should cover this in a UX review.

src/app/search/directives/search-series.component.ts
cavis: Separate PR it is!

src/app/search/directives/story-list.component.ts
```diff
@@ -11,14 +11,24 @@ import { StoryModel } from '../../shared';
...
   @Input() noStories: boolean;
   @Input() stories: StoryModel[];
   @Input() loaders: boolean[];
+
+  emptyCards: number[];
+
+  ngOnChanges(changes: any) {
+    if (changes['stories'] && changes['stories'].currentValue &&
+        changes['stories'].currentValue.length > 0 && changes['stories'].currentValue.length % 3 > 0) {
+      this.emptyCards = Array.from(Array(3 - changes['stories'].currentValue.length % 3), (x, i ) => i);
```
cavis: What happens when there are 3 results, but your viewport is narrowed down to 2-wide?

sandikbarr: The extra empty card gets wrapped but the zero height doesn't affect layout. Really glad you thought of this zero height empty cards idea because I did not want to handle component width/height. I don't know about viewport, but I tried checking component width or height for something else and it always came back zero.

cavis:
👍 :ship: :it:

PR #51, opened 2016-09-30
Series image upload

cavis:
Implement an image uploader for the series-edit page.

- Abstracts tabs into a shared component between series/story
- Allow image-uploader component to work for either story (has-many) or series (has-one)
- Series deletion button (not working yet)
- Bonus: global error handler


cavis:
@sandikbarr i think that PRX/cms.prx.org#169 may change this a bit - but I'll revisit after that merges.  There also seems to be an issue with DELETEs not actually deleting the series - but that will also have to be addressed in CMS.

My specs for the series-tabs are pretty lame.  But I think it's a good idea to at least make sure each component renders without errors, even if there's no non-core functionality to test.

src/app/series/series.routing.ts
```diff
@@ -3,22 +3,42 @@ import { Routes, RouterModule } from '@angular/router';
...
+    children: [
+      { path: '',          component: SeriesBasicComponent },
+      { path: 'image',     component: SeriesImageComponent },
+      { path: 'templates', component: SeriesTemplatesComponent },
+      { path: 'advanced',  component: SeriesAdvancedComponent }
+    ]
   },
   {
-    path: 'series/:id',
+    path: 'series/new',
```
sandikbarr: Should this route be above :id since it's more specific? Looks like you may not even need the new route as a separate route since the route configs are basically the same.

src/assets/styles/forms.css
```diff
@@ -168,6 +168,14 @@ button.btn-link.active {
   color: #f59f51;
 }
 
+button.delete {
+  color: #fff;
+  background-color: #ee4d3e;
+}
+button.delete:hover {
```
sandikbarr: probably super nit picky, but maybe we should add :focus here too

src/app/core/modal/modal.component.ts
```diff
@@ -36,6 +37,7 @@ export class ModalComponent {
         this.setScroll(false);
       }
       this.state = state;
+      ref.detectChanges();
```
cavis: I think I added that specifically for the error-handling case.  A top-level error is caught without any zone changes... so the modal wouldn't appear with the stack trace until something else happened.

There may be a cleaner way to do this, only in the error-handling case.  Will look into it.

src/app/app-error.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    this.defaultHandler = new ErrorHandler(false); // don't re-throw
+  }
+
+  handleError(err) {
+    this.modal.show({
+      title: 'Uncaught ' + (err.name || 'Error'),
+      body: `
+        <p>${err.message}</p>
+        <hr/>
+        <pre style="font-size:12px;line-height:16px;white-space:pre">${err.stack}</pre>
```
cavis: I was having some problems getting any `modal.component.css` to apply to the actual contents rendered in the modal.  Since angular is doing the whole "emulated shadow dom" thing, all the styles get a bunch of extra attribute-selectors added at runtime, which my `body:` does not include.

Any other ideas?  Could also revisit this when we get even better error handling.  (Prod-vs-dev errors, links to related error data, etc).

src/app/series/series.routing.ts
cavis: Good point - I hadn't noticed if order-of-routes mattered.  Will double check that.

I do like having an explicit `series/new` route - so I don't have to check if `params.id === 'new'` when the component subscribes to the route params.  Maybe I can just DRY this up by putting the child routes in a shared const or something.

src/assets/styles/forms.css
cavis: Form'd!

cavis:
Per the delete-before-confirmation ... yeah, wondering if the modal isn't binding/unbinding to the key events correctly.  So it grabs the first key event (to open the modal) before it's even rendered.

src/app/app-error.service.ts
sandikbarr: Ew, I see what you mean, and now I like this pseudo template even less. :( That's not very helpful, but I will keep this in mind and think about potentially using a different approach.

src/app/core/modal/modal.component.ts
cavis: Tried a couple other things (complete manual change detection, triggering from the error handler instead of here), and I think this is still the cleanest.  Since this component can trigger change-detection on only itself (and children), it shouldn't cost anything to have an extra `detectChanges()` in there, if this modal was triggered by a normal angular zone change.

src/app/app-error.service.ts
cavis: Fixed!  (By using the deep `>>>` css selector).

cavis:
Fixed delete-too-early problem by listening to `keydown` instead of `keyup`.  It was catching the same keypress as the button that opened the modal, which immediately clicks "Okay".

Issue #55, opened 2016-09-30
search state should be reflected in URL


sandikbarr:
Include the following search parameters in the browser URL with [optional route parameters](https://angular.io/docs/ts/latest/guide/router.html#!#optional-route-parameters)
- series id (on stories when searching by series)
- search text
- paging/view all
- order by and direction
- genre/sub genre
- (eventually) tags

Background: Story search currently supports an optional seriesId URL parameter to link directly into search by series from the dashboard. If the user changes the dropdown after navigating to search by series, the selected seriesId is not reflected in the URL. So the seriesId search parameter should always be included in the URL when set and updated when changed.

PR #70, opened 2016-10-12
Search parameters in the URL

sandikbarr:
Adds search and paging parameters to the URL as defined in issue #55


src/app/search/search.component.ts
```diff
@@ -20,152 +22,105 @@ export class SearchComponent implements OnInit {
...
-  search(page: number, per: number = undefined) {
-    if (this.isOnSeriesTab()) {
-      this.searchSeries(page, per);
-    } else if (this.isOnStoriesTab) {
-      this.searchStories(page, per);
+  navigateToPagePer(page: number, perPage: number = undefined) {
+    if (this.selectedTab === SearchComponent.TAB_STORIES) {
+      this.searchWithParams(Object.assign({}, this.searchStoryParams, {tab: this.selectedTab, page, perPage}));
+    } else if (this.selectedTab === SearchComponent.TAB_SERIES) {
+      this.searchWithParams(Object.assign({}, this.searchSeriesParams, {tab: this.selectedTab, page, perPage}));
```
sandikbarr: I was worried that I might need a polyfill for Object (we did on a Babel project) but typescript takes care of it. I miss having the spread operator for objects though, which I have a feeling will make me sad when doing Redux.

src/app/search/search.component.ts
```diff
@@ -20,152 +22,105 @@ export class SearchComponent implements OnInit {
...
+      this.auth = auth;
+
+      this.allSeriesIds = [-1];
+      this.allSeries = {};
+      this.auth.followItems('prx:series', {filters: 'v4'}).subscribe((series) => {
+        for (let s of series) {
+          this.allSeriesIds.push(s.id);
+          this.allSeries[s.id] = new SeriesModel(this.auth, s, false);
         }
+        this.subscribeRouteParams();
```
cavis: Why wait for the CMS requests to parse the route params?

src/app/search/directives/search-series-form.component.css
```diff
@@ -1,3 +1,7 @@
+form {
+  display: block;
```
sandikbarr: No, forms.css makes them display:flex. I haven't verified that's actually being used though. This file needs to go away eventually though as it's very generic styling that belongs in the styleguide.

src/app/search/search.component.ts
sandikbarr: The Observable on the route params parses the route params then calls the search functions which need the auth and/or series to make requests on.

sandikbarr: sounds like we need to enter an issue for this!

PR #72, opened 2016-10-12
Reorganize and improve story tabs

cavis:
- Reorganize story-tab content
- Remove genre/subgenre, and replace with plain ol' story-tags
- Restyle story-edit header to be thinner (same height as the series-edit header)
- After you save an audio/image, poll CMS until the status changes from `uploaded` to `complete`.  (Or some error status).
- Fix a lot of UX issues with the uploaders


src/app/home/home.component.html
```diff
@@ -2,7 +2,6 @@
   <section>
     <header>
       <h2>Your Series</h2>
-      <a class="create" [routerLink]="['/series/new']">New Series &raquo;</a>
```
cavis: This link was unnecessary and ugly, as there's already a "Create Series" link in the header.

src/app/series/series.component.html
```diff
@@ -10,7 +10,7 @@ <h1 class="hero-title" *ngIf="!id">Create Series</h1>
 
     <template [ngIf]="series.isNew">
       <publish-button [model]="series" plain=1 working=0 disabled=0 (click)="discard()">Discard</publish-button>
-      <publish-button [model]="series" visible=1 orange=1 (click)="save()">Create</publish-button>
+      <publish-button [model]="series" visible=1 green=1 (click)="save()">Create</publish-button>
```
cavis: Making the create buttons green calls them out a bit more.

src/app/shared/model/audio-file.model.ts
```diff
@@ -49,7 +61,11 @@ export class AudioFileModel extends UploadableModel {
   }
 
   saveNew(data: {}): Observable<HalDoc> {
-    return this.parent.create('prx:audio', {}, data);
+    return this.parent.create('prx:audio', {}, data).map(doc => {
+      this.setState();
+      this.watchProcess();
```
cavis: This starts polling-for-status-change as soon as you save a new file.

Issue #53, opened 2016-09-30
search cards should clear when search changes and loaders count is incorrect


sandikbarr:
Currently search is attempting to get a count of results from the parent, but that count does not take into account the filter parameters to be used in the request. So a bunch of extra loaders are shown when doing a search.

On load and when the search changes
- [ ] remove all the current cards
- [ ] show a single blank-card with a spinner on it

PR #74, opened 2016-10-13
Fix search loaders

sandikbarr:
#53 
- Loading status is now a single card with a loading spinner
- Search results are cleared when new search parameters are given


src/app/search/directives/story-list.component.ts
```diff
@@ -21,7 +21,7 @@ import { StoryModel } from '../../shared';
 export class StoryListComponent {
   @Input() noStories: boolean;
   @Input() stories: StoryModel[];
-  @Input() loaders: boolean[];
+  @Input() isLoaded: boolean[];
```
cavis: Array?

Issue #52, opened 2016-09-30
provide a means to clear select and show placeholders (select2 style)


sandikbarr:
In search, the search parameters implemented with a select do not show a placeholder and do not allow the user to easily clear the search selection for the series or genre filters. Publish should provide a means for the user to clear those selections, perhaps through a placeholder/empty option or by using a select2 implementation that supports a placeholder value and provides an `X` clear button.

PR #77, opened 2016-10-14
provide a means to clear search select filters

sandikbarr:
I don't particularly like any of our options right now, but this seems like it would give us the least amount of trouble in the meantime. If we can settle on this for an interim solution for #52 then I'll Icebox an issue for the select2 dropdown style. (The Angular2 story on Chosen is not any better, BTW.) I could potentially put the X button inside the select box with a negative margin if we think that would be better. I could also not disable the placeholder option and get rid of the X button. Let me know you opinions if you've got any.

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -61,6 +61,7 @@ export class SearchComponent implements OnInit {
 
   subscribeRouteParams() {
     this.route.params.forEach((params) => {
+      this.clearResults();
       this.selectedTab = params['tab'] || SearchComponent.TAB_STORIES;
       this.currentPage = params['page'] ? +params['page'] : 1;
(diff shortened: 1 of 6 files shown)


src/app/search/search.component.ts
```diff
@@ -74,10 +75,9 @@ export class SearchComponent implements OnInit {
   }
 
   searchWithParams(searchParams: any) {
-    this.clearResults();
     let routeParams = {tab: this.selectedTab};
     Object.keys(searchParams).forEach(key => {
-      if (searchParams[key] !== undefined) {
+      if (searchParams[key] !== undefined && searchParams[key] !== 'undefined') {
```
sandikbarr: ngModel matching needs the disabled option value to be "undefined" rather than empty string

Issue #63, opened 2016-10-07
Add audio templates to story edit view for uploads


kookster:
Have a guide or other hinting of the number, label, and duration of audio to be added to an episode based on the previously set up audio templates for a series/podcast

PR #79, opened 2016-10-20
Audio file/version templates

cavis:
See #63.

- When creating a new story with a series:
 - Grab the audio-version-template for the series (if it exists)
 - Grab any audio-file-templates for that audio-version-template
 - Validate number-of-segments and audio duration based on that
- When editing an existing story:
 - Grab the audio-version-templates for any audio-versions in that story (if they exist)
- If no templates:
 - Validate that the audio-version has at least 1 file
 - Validate that any audio-file is actually audio (for now it just uses HTML5 Audio to determine if it's "playable")

Also spent some time fighting change detection, and making sure that validation messages are displayed in a sane way.


cavis:
@sandikbarr I'm still trying to track down a couple UX bugs.  But I most likely introduced them before this PR ... so have at!

PR #81, opened 2016-10-21
better placeholder images

sandikbarr:
Some images that makes a little more sense. I could also make these SVGs and update the filenames in the code if that is preferred. I went with PNG since that is what we had, but that was because I was only using Preview to create the images at the time.


cavis:
👍

PR #84, opened 2016-11-04
Edit audio templates

cavis:
Ability to edit audio version/file templates on the series.

- [x] Add/remove audio-version-templates on the series
- [x] Allow a maximum of 1 per series for now (since we're only doing the podcast audio-version)
- [x] Add/remove audio-file-templates within the audio-version-template

Makes no attempt to warn the user about the "bad things" that might happen if you change the templates for a series with existing stories.  We'll probably lock that down more at some point in the future - but this PR at least gets all the functionality in place.

### DIFF
--- a/src/app/series/directives/series-templates.component.ts
+++ b/src/app/series/directives/series-templates.component.ts
@@ -1,32 +1,53 @@
 import { Component, OnDestroy } from '@angular/core';
 import { Subscription } from 'rxjs';
-import { SeriesModel, TabService } from '../../shared';
+import {
+  SeriesModel,
+  TabService,
+  AudioVersionTemplateModel,
+  AudioFileTemplateModel
+} from '../../shared';
 
 @Component({
   styleUrls: ['series-templates.component.css'],
   template: `
     <form *ngIf="series">
 
-      <div *ngFor="let v of series.versionTemplates" class="version">
-
-        <publish-fancy-field textinput required [model]="v" name="label" label="Version Label">
-        </publish-fancy-field>
-
-        <publish-fancy-field class="length" [model]="v" label="Total length in seconds" invalid="lengthAny">
-          <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMinimum" label="Minimum">
+      <template ngFor let-v [ngForOf]="series.versionTemplates">
+        <div *ngIf="!v.isDestroy" class="version">
+          <publish-fancy-field textinput required [model]="v" name="label" label="Version Label">
+            <div class="actions">
+              <a (click)="removeVersion(v)"><i class="icon-cancel"></i>Remove Template</a>
+            </div>
           </publish-fancy-field>
-          <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMaximum" label="Maximum">
+
+          <publish-fancy-field class="length" [model]="v" label="Total length in seconds" invalid="lengthAny">
+            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMinimum" label="Minimum">
+            </publish-fancy-field>
+            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMaximum" label="Maximum">
+            </publish-fancy-field>
           </publish-fancy-field>
-        </publish-fancy-field>
 
-        <publish-fancy-field label="Audio Segments">
-          <publish-file-template *ngFor="let t of v.fileTemplates" [file]="t">
-          </publish-file-template>
-          <div *ngIf="!v.fileTemplates.length" class="empty">
-            <h4>No segments defined</h4>
-          </div>
-        </publish-fancy-field>
+          <publish-fancy-field label="Audio Segments">
+            <div class="actions">
+              <a *ngIf="canAddFile(v)" (click)="addFile(v)"><i class="icon-plus"></i>Add Segment</a>
+              <a *ngIf="canRemoveFile(v)" (click)="removeFile(v)"><i class="icon-cancel"></i>Remove Segment</a>
+            </div>
+            <publish-file-template *ngFor="let t of v.fileTemplates" [file]="t">
+            </publish-file-template>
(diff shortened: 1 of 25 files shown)


src/app/shared/form/fancy-field.component.html
```diff
@@ -1,4 +1,4 @@
-<div [class]="fieldClasses" [class.small]="small">
+<div [class]="fieldClasses">
```
cavis: This was a bug - any change to the `fieldClasses` after the component's initial render would actually overwrite `class.small`.

src/app/shared/form/fancy-field.component.ts
```diff
@@ -15,15 +17,36 @@ export class FancyFieldComponent {
...
+  @Input()
+  set select(opts: string[]) { this.type = 'select'; this._select = opts || []; }
+  get select() { return this._select; }
+
+  // Field attributes
+  _small = false;
+  _inline = false;
+  _required = null;
+  @Input()
+  set small(small: boolean) { this._small = isset(small); }
```
cavis: These interceptors allow a simpler syntax for using this component ... instead of doing `small="true"` in the template, you can just use the standalone attribute `small`.

src/app/shared/model/audio-version-template.model.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  invalid(field?: string | string[]): string {
+    if (field === 'lengthAny') {
+      return this.invalid('lengthMinimum') || this.invalid('lengthMaximum');
+    } else {
+      return super.invalid(field);
+    }
+  }
+
+  findUnsavedFiles(position, found: AudioFileTemplateModel[] = []) {
```
cavis: This was a bit tricky ... any unsaved file-templates get stored with their `position` in the `key()`.  So this tries to recursively find any unsaved file-templates in localstorage.

src/app/shared/model/base.model.ts
```diff
@@ -118,7 +118,12 @@ export abstract class BaseModel {
       } else {
         model.parent = this.doc;
       }
-      return model.save();
+      return model.save().map(saved => {
+        if (saved && model.isDestroy) {
+          this.removeRelated(model);
```
cavis: Related models weren't getting removed from the array after they were DELETEd.

src/app/shared/model/base.model.ts
```diff
@@ -137,6 +142,11 @@ export abstract class BaseModel {
     this.unstore();
     this.lastStored = null;
     this.isDestroy = false;
+    if (!this.doc && this.original) {
```
cavis: This is a bit of a typescript gotcha.  All the default attributes at the top of the class (like `public thing = 'default-value-here';`) are actually evaluated as part of the constructor.  So for unsaved models, I need to recover those values from the `this.original` hash.

src/app/shared/model/invalid/template.invalid.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (min === null && max === null) {
+    return null; // allow unset
+  } else if (min === null) {
+    return `Must set a minimum`;
+  } else if (isNaN(parseInt(min, 10))) {
+    return `Minimum is not a number`;
+  } else if (min < 1) {
+    return `Minimum must be greater than 0`;
+  } else if (min >= max) {
+    return 'Minimum must be less than maximum';
```
sandikbarr: Feel free to ignore this, but I wondered if we could not show the invalid message when the user is filling out the form in order and hasn't entered a maximum yet. Looks like they can have both be unset but if they set one, they have to set both. Is that just because that's how templates work? What if they want a min but no max?

src/app/series/directives/series-templates.component.ts
```diff
@@ -1,11 +1,56 @@
...
+            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMinimum" label="Minimum">
+            </publish-fancy-field>
+            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMaximum" label="Maximum">
+            </publish-fancy-field>
+          </publish-fancy-field>
+
+          <publish-fancy-field label="Audio Segments">
+            <div class="actions">
+              <a *ngIf="canAddFile(v)" (click)="addFile(v)"><i class="icon-plus"></i>Add Segment</a>
+              <a *ngIf="canRemoveFile(v)" (click)="removeFile(v)"><i class="icon-cancel"></i>Remove Segment</a>
```
cavis: I'd like to revisit after I do this audio-uploader-revamp PR.  It's sort of problematic to go changing the number of segments after you've started creating stories against that template.  So this template-editor may end up being more of a "template-creator-only".  Or maybe you can change the durations/labels after creation, but not the number of segments.

src/app/series/directives/series-templates.component.ts
```diff
@@ -22,4 +67,54 @@ export class SeriesTemplatesComponent implements OnDestroy {
...
+  removeVersion(version: AudioVersionTemplateModel) {
+    version.isDestroy = true;
+    if (version.isNew) {
+      this.series.versionTemplates.splice(this.series.versionTemplates.indexOf(version), 1);
+      version.unstore();
+    }
+  }
+
+  canAddFile(version: AudioVersionTemplateModel): boolean {
+    return version.fileTemplates.filter(f => !f.isDestroy).length < 10;
```
cavis: Arbitrary ... same way I determined the min-length validation for title/description fields.  Ticket #86 is to actually find some sane values for validations.

src/app/shared/model/invalid/template.invalid.ts
cavis: That's currently how the server validation works - if either of them are set, both must be set.  I'm also guessing that most users will fill in both numbers, since they'll have a pretty good idea of sane lengths for their show.

If that's not the case, could do something like send a really-big-number if max is missing (and send 0 if min is missing).  Or... change the server code.

Issue #96, opened 2016-11-04
Audio uploader redesign


cavis:
Split audio uploader into 2 experiences:

1. Non-templated audio
    - [x] Keep the current single-button multi-file uploader
    - [x] Keep the current drag-and-drop reordering
    - [x] Allow labeling the segments, but ditch the dropdown-triangle
2. Templated audio
    - [x] Display an upload button next to each segment (single-file)
    - [x] Allow user to delete/replace existing files
    - [x] Changing labels not allowed (they come from the template)

Additionally:

- [x] Add a single-segment template to a new series by default

PR #98, opened 2016-11-14
Revamp audio uploader

cavis:
See #96.  Support 1-per-line file inputs for templated stories, while continuing to use the single multi-file-input for untemplated stories.

- Refactored `shared/upload` into:
  - a top level `upload` module for audio uploader (the bulk of the code)
  - a `shared/image/image-upload.component` for the simpler (and actually shared) image uploader
- Generally broke the old `audio-file.component` into smaller components/directives


src/app/upload/shared/audio-cancel.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  @Input() version: AudioVersionModel;
+
+  @HostListener('click') onClick() {
+    this.publishAudioCancel.canceled = true;
+    setTimeout(() => {
+      this.publishAudioCancel.destroy();
+      this.version.removeUpload(this.publishAudioCancel);
+      this.publishAudioCancel.canceled = false;
+    }, this.delay);
```
sandikbarr: This transition is a very nice touch. :) Also, hmm so that's how defaults are done, makes so much sense (still so much to learn, big and small.)

Issue #95, opened 2016-11-04
Duration validation doesn't happen until upload completes


cavis:
We know the length of the mp3 as soon as it's selected, so why wait for the upload to finish before telling the user they're wrong?

Issue #65, opened 2016-10-07
Add playback of audio to be uploaded and saved to a story


kookster:

cavis:
Fallout: seeking is not entirely bug-free PRX/publish.prx.org#102

Issue #62, opened 2016-10-07
Client side audio validations


kookster:
At least length, audio type, potentially encoding such as sample rate and bitrate

PR #99, opened 2016-11-18
Audio playback and validation

cavis:
See #62 and #65 and #95.

Requires PRX/cms.prx.org#189

- Add `aurora.js` and `mp3.js` as global includes (didn't make any attempt to just-in-time load them)
- Validate files via aurora (gives an exact duration, and file type)
- Playback uploading/uploaded files via aurora (need that CMS PR for CORS reasons)
- New express server for dev and dist
  - Intercepts requests for `index.html` and manually renders the pug template, including the ENV
  - For dev, proxy file requests to a spawned `ng serve` process on a different port
  - For dist, statically serve files in the `/dist` directory
  - Proxy `/pub` requests for CMS, to overcome an issue with browser not following redirects correctly (very recent changes to the CORS spec should fix this, but not waiting for that)


lib/util.js
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Find the script tags to include
+ */
+exports.findScripts = (isDist) => {
+  let scripts = ['inline.js'];
+  if (isDist) {
+    let files = [];
+    try { files = fs.readdirSync(`${__dirname}/../dist`); } catch (e) {}
+    scripts.push(files.find(f => f.match(/^styles\..+\.bundle\.js$/)));
```
cavis: The built `/dist` files have a hash string in them ... so i just scanned the directory to match on them.

src/app/core/audio/player.service.ts
```diff
@@ -0,0 +1,101 @@
...
+      let data = <AudioMetadata> {};
+      let asset = AV.Asset.fromFile(file);
+      let update = () => {
+        if (asset.duration) {
+          data.duration = Math.round(asset.duration / 1000);
+        }
+        if (asset.format) {
+          data.format = asset.format.formatID;
+          // TODO: format doesn't actually include data, so this never works
+          if (data.format === 'mp3' && asset.format.layer === 2) {
```
cavis: This code came from the www.prx.org project ... but doesn't seem to work here.  Maybe I have the wrong version/extensions of Aurora?  You'd think mp3.js would know this, since it's adding support for mpeg layer 2.

src/app/core/audio/player.service.ts
```diff
@@ -0,0 +1,101 @@
...
+        sub.next(data);
+      });
+      player.on('end', () => sub.complete());
+      player.on('error', err => sub.error(err));
+      player.play();
+
+      this.playingPlayer = player;
+      this.playingData = data;
+
+      return () => player.stop();
```
cavis: This actually worked out pretty well - whenever the caller unsubscribes from this playback observable, the player will stop.

src/app/core/audio/player.service.ts
```diff
@@ -0,0 +1,101 @@
...
+  duration: number;
+  progress: number;
+}
+
+@Injectable()
+export class PlayerService {
+
+  private playing: Subscriber<PlaybackMetadata>;
+  private playingPlayer: AV.Player;
+  private playingData: PlaybackMetadata;
```
cavis: A bit wonky, but I added these "currently playing" singletons to support the static `stop()` and `seek()` methods on this service.

src/app/core/upload/upload.service.ts
```diff
@@ -73,7 +72,6 @@ export class Upload {
     this.uuid = UUID.UUID();
     this.name = file.name;
     this.size = file.size;
-    this.duration = file['duration'];
```
cavis: Instead of setting the duration on the `Upload` and `File` ... it's now set on the `AudioFileModel` along with bitrate/frequency/etc.  Which makes way more sense.

angular-cli.json
```diff
@@ -17,7 +17,10 @@
       "styles": [
         "styles.css"
       ],
-      "scripts": [],
+      "scripts": [
```
sandikbarr: I feel like you know my pain. You can include scripts from node_modules here too. So might be able to do 
`scripts: ["../node_modules/av/src/aurora.js", "../node_modules/mp3/whatever`

src/app/upload/shared/audio-player.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    <button *ngIf="!playing && !loading" class="play" (click)="play()"></button>
+
+    <div #scrubber *ngIf="playing" class="scrubber"
+      (mousedown)="scrub('down', $event)"
+      (mouseup)="scrub('up', $event)"
+      (mousemove)="scrub('move', $event)">
+      <div class="meter" [style.width.%]="progress * 100"></div>
+      <div *ngIf="!dragging" class="position" [style.left.%]="progress * 100"></div>
+    </div>
+    <button *ngIf="playing" class="pause" (click)="stop()"></button>
```
sandikbarr: I'm just being picky here, but this is more of a stop button because it doesn't pause and allow you to restart the audio from the same position, seems kinda confusing.

src/app/upload/shared/audio-player.component.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.sub.unsubscribe();
+    }
+    this.loading = false;
+    this.playing = false;
+    this.progress = 0;
+    this.ref.detectChanges(); // playback 'done' doesn't fire by default
+  }
+
+  scrub(dir, ev: MouseEvent) {
+    let pct = Math.max((ev.offsetX - 2) / this.scrubberWidth, 0);
```
sandikbarr: Scrubbing isn't working for me. When I move the scrubber, the audio is jerky briefly then seems to stop playing. The pause button still shows as if it should be playing, but it doesn't progress any further.

cavis: I can attempt to recreate - what browser are you using?  And what duration is the file?

src/app/upload/shared/audio-player.component.ts
cavis: Valid point.  Can definitely maintain the position instead of starting over at 0.

What functionality would you expect if you clicked play on a different file (which will "pause" the first file), and then go back to the first file again and click play?  Starts at beginning?  Or where you last left off?

src/app/upload/shared/audio-player.component.ts
cavis: Also, are you trying to drag/drop the scrubber?  Or just clicking on the point you want to seek to?  (Do either of those work?)

sandikbarr: Same behavior whether I drag it or if I click further in the progress bar. Chrome Version 54.0.2840.71 (64-bit). Happens on both mp3 files I often use for testing, one is 4 seconds, the other is 5:55.

src/app/upload/shared/audio-player.component.ts
sandikbarr: If the progess can be saved on the other file without a ton of work, pausing would be nice, but maybe a "nice to have" but not necessary. Restarting the other file seems like it would be fine too.

src/app/upload/shared/audio-player.component.ts
cavis: Okay, i _think_ i've tracked-down-but-not-fixed the issue.  Seems that Aurora does nothing to ensure you don't seek past the duration it's streamed so far.  And if you do seek past the end, it throws an error.  I need to add some logic to wait for enough to download before playing.

I am a little confused by the 4-second file exhibiting this problem.  I've had problems with playing a 1-second file ... but only because the playback sometimes completes before the initial `progress` event fires.

sandikbarr: bleh, maybe we should just merge this and open a bug ticket

Issue #91, opened 2016-11-04
Incorporate embed player


cavis:
Add the embed player to the story-embed tab.  Pre-fill all fields, show a preview, and just give the user a copyable link to use.

Bonus points: allow user to alter some aspects of the player - but don't save changes to CMS or anything crazy.  Just change the copyable link.

gcampo88:
Working on this; one question for @cavis / anyone: I can pull the title out of the StoryModel object, but currently there's no way to get the audio url with ads, etc, correct? Is that something that will be doable soon? I'm wondering if I should flash a message to the user that they need to save their Story before they can see the player preview, since it'll rely on the feed URL, image URL, audio URL etc which will only be set up stably once they hit Save, right? Holler if I'm way off base...

cavis:
You're right - CMS currently doesn't know anything about the destination podcast feed.  (This is related to what @kookster is currently working on).  You're also not guaranteed to have a single mp3 file - it may be multiple files at this point.

I'm inclined to say "fake it" for what you're missing, as we're most interested in feedback over functionality.  Meaning ... plug in the CMS-returned values for the image and first-audio-file, and leave the subscription-url blank.  (Unless the player already supports multiple file inputs?)

cavis:
Also, sticking with the UX themes of Publish ... maybe the embed player you show on the page (the "preview") doesn't need to be the link you give them to copy.  Meaning, you always want them to copy a URL using the currently-live feed, but you want to show them what pending changes "would look like eventually" after they click save.

gcampo88:
not sure what you mean with the "currently-live feed" -- if this is a new story there's no live feed at the moment, right? Or do you just mean the first audio file?

cavis:
So eventually, CMS will be able to tell for any story whether it has an RSS feed distribution, and what the URL to that is.  If it exists, make sure the copyable-player-link you give them is for the public feed, which could be out-of-sync with the data CMS has.

But the preview of the player you should them should pull from CMS (or maybe even unsaved localstorage changes), with the latest data.

And if the story doesn't have an RSS feed ... well, that's another good question.  We could give them a non-podcast-audio player, that just links to the CMS audio files stored in S3.  Or we can just not give them anything, and say only podcasts get the player.

Issue #87, opened 2016-11-04
Word-smithing


cavis:
Clarify field labels and hints.  Specifically:

- [ ] Story audio-versions - what are they?  How to label when only 1?
- [x] Story image - where is this used?  What should it be?
- [x] Story explicit - link to itunes docs.  And use "Explicit/Clean" instead of "Yes/Clean".
- [x] What are "Standalone Stories" (on the dashboard)?
- [x] What's the difference between "Draft" and "Unpublished" stories?
- [x] What does "Stored Locally" mean, when you try to leave a draft story?
- [x] Story tags - where do these go?
- [x] Series title - maybe more like "What's the name of your series?"
- [x] Series description - better hint

Generally, some confusion about how the "Basic" story edit page relates to podcast information.  (Podcast specific info should be tab 2, but that's non-obvious).  Things feel "done" when the first tab is done - what motivates someone to go to tab 2?

Issue #86, opened 2016-11-04
Finalize validations


cavis:
Not sure exactly what these should be yet, but should definitely think about them

- [x] Story title/teaser/description min-length
- [x] Series title/teaser/description min-length
- [x] Audio template label min-length

PR #101, opened 2016-11-18
Feat/add embed player -- WIP

gcampo88:
- [X] Addressed #86 by making story + series title and teaser validations required only, no min length; left length validation (10 char) on description fields for now. 
- [X] Addressed #87 to clarify phrasing on Edited/Draft/Unpublished/Published status, "saved locally" message wording, explicit/clean, etc. 
- [X] Created :hover tooltips to show extra information if we want them, though I think I actually would vote to just have all hints be in the `<div class="fancy-hint">` hint areas to keep it cleaner. I've currently left some tooltips in place with "tbd" text just to show what they'd look like-- we should clear out that text before merging. 
- [X] Addressed #91 by adding iframe to hold piece embed player; using mostly placeholder values for now as CMS-Feeder integration work continues.


src/app/home/directives/home-series.component.ts
```diff
@@ -9,6 +9,9 @@ import { StoryModel } from '../../shared';
   styleUrls: ['home-series.component.css'],
   template: `
     <header *ngIf="noseries">
+      <section [publishHoverTip]="'These are your stories not associated with any Series.'" class="hover-tip">
```
cavis: Could add the `hover-tip` class by using a `HostBinding('class.hover-tip') = true` in the directive, and cut down on some duplication.

If those hover tip strings get too long (enough to frighten the linter), you could start moving those to component properties.

src/app/home/directives/home-story.component.ts
```diff
@@ -88,10 +88,10 @@ export class HomeStoryComponent implements OnInit {
       this.statusText = 'Draft';
     } else if (this.story.changed()) {
       this.statusClass = 'status unsaved';
-      this.statusText = 'Edited';
+      this.statusText = 'Unsaved Changes';
     } else if (!this.story.publishedAt) {
       this.statusClass = 'status unpublished';
-      this.statusText = 'Unpublished';
+      this.statusText = 'Private';
```
cavis: Yesss!  (I'm a fan of calling things public/private instead of published/unpublished).  Eventually we'll also want to move the "make private again" button somewhere less prominent.  But that's another UX problem for another day.

src/app/shared/model/story.model.ts
```diff
@@ -22,9 +22,10 @@ export class StoryModel extends BaseModel {
   SETABLE = ['title', 'shortDescription', 'description', 'tags'];
 
   VALIDATORS = {
-    title:            [REQUIRED(), LENGTH(10)],
-    shortDescription: [REQUIRED(), LENGTH(10)],
-    description:      [LENGTH(10)]
+    title:            [REQUIRED()],
+    shortDescription: [REQUIRED()],
+    description:      [LENGTH(10)],
+    tags:             [REQUIRED()]
```
cavis: Are tags really required now?

src/app/shared/shared.module.ts
```diff
@@ -26,7 +28,8 @@ import { TabComponent } from './tab';
     ImageUploadComponent,
     SpinnerComponent,
     TabComponent,
-    TimeAgoPipe
+    TimeAgoPipe,
+    HoverDirective
```
cavis: This is probably OCD Ryan talking ... but i was aiming for keeping the imports/exports sorted alphabetically in these module files.

src/app/story/directives/basic.component.ts
```diff
@@ -7,15 +7,15 @@ import { StoryModel, TabService } from '../../shared';
   template: `
     <form *ngIf="story">
 
-      <publish-fancy-field [model]="story" textinput="true" name="title" label="Story Title" required>
+      <publish-fancy-field [model]="story" textinput="true" name="title" label="Story Title" tiptext="This will show up in the \'Name\' column on iTunes." required>
```
cavis: Alternatively, you could create a `class="fancy-tip"` element inside the field, the same way hints work.  And then pull that selector out in the component - which could also allow you to use html tags in the tooltip.  Okay, so maybe that's not all that useful.

src/app/shared/model/story.model.ts
gcampo88: Nope, my mistake! I assumed these would double as iTunes categories, which are required for podcasts -- but AK disabused me of that notion.

src/app/shared/shared.module.ts
gcampo88: 👍

src/app/home/directives/home-story.component.ts
gcampo88: 👍

src/app/home/directives/home-series.component.ts
gcampo88: Cool, thanks! Docs for HostBinding weren't great -- went with `@HostBinding('attr.class') class = 'hover-tip';`. That ok?

cavis: The downside is that would replace any classes defined on the parent element, rather than adding to them.  I think there are some `class.something` hostbindings in Publish somewhere already.  I ended up getting examples from [stackoverflow](http://stackoverflow.com/questions/37965647/hostbinding-and-hostlistening-what-do-they-do-and-what-are-they-for), since the docs just aren't there yet.

gcampo88: Good call. Experimented a bit, looks like this solves it:  `@HostBinding('class.hover-tip') true;`

gcampo88:
@cavis quick question: I did the tooltip through an attribute directive, but this feels like it would be more natural to do purely through CSS -- is there a reason to avoid doing so?

cavis:
👍 looks good to me.  At least worth some feedback on those "tbd"s

src/app/story/directives/player.component.ts
```diff
@@ -1,49 +1,122 @@
...
+      let img = this.story.images.find(img => !img.isDestroy);
+      if (img) { this.imageUrl = img.enclosureHref; }
+    }
+
+    if (this.story.versions.length > 0) {
+      let firstAudio = this.story.versions[0].files.find(file => !file.isDestroy);
+      if (firstAudio) { this.audioUrl = firstAudio.enclosureHref; }
+    }
+
+    if (this.story.doc && this.story.doc._links['prx:series']) {
```
gcampo88: :+1:

src/app/story/directives/player.component.ts
```diff
@@ -1,49 +1,122 @@
...
+  iframeHtml(width: string, height: string) {
+    const url = this.embeddableUrl;
+    return `<iframe frameborder="0" height="${height}" width="${width}" src="${url}"></iframe>`;
+  }
+
+  copy(el: any) {
+    const sel = el.dataset.copytarget;
+    const inp = document.querySelector(sel);
+    if (inp && inp.select) {
+      inp.select();
```
gcampo88: cool, thanks!

gcampo88:
back to you @sandikbarr !

Issue #47, opened 2016-09-21
Use a wysiwyg for story full description


kookster:
Previously a wysiwyg js editor was used that worked with markdown as the saved format of the description text.

This was to allow some but not unrestricted formatting, and to make it possible to return both html and plain text versions of the description in addition to markdown with relative ease.

As we move to have the CMS publish stories to producer websites and other types of posts, having a more rich but limited form of formatting with inclusion of links and images makes sense to provide.
- [x] Select and import a wysiwyg js editor, adding ng2 module or other needed integration code
- [x] Integrate the editor into the publish page for the full description
- [x] Customize the editor look and feel, and formatting options to basics (i.e. h# tags, p tags, links, images, italic, bold)
- [x] Add integration of images uploaded to the story placed into the description text

kookster:
If markdown is selected for the saved format, has repercussions for the CMS api, and format of V3 content - perhaps requiring a parallel field for v3 continuity

Issue #88, opened 2016-11-04
Image uploader not replacing


cavis:
- [x] Replacing image doesn't remove the old one from the page right away.
- [x] No "Changed" indicator after you remove an image.  (Maybe highlight the uploader box in orange?)

cavis:
First issue was fixed awhile back - it was a problem with the localstorage copy of the deleted image not getting cleaned up.

PR #103, opened 2016-11-29
WYSIWYG editor for story description markdown formatting

sandikbarr:
#47 Use a wysiwyg for story full description. Includes the ability to add headings, paragraphs, strong or italic text, links to user provided URLs, images from the story, as well as lists, HRs, and preformatted/code style bonus features


src/app/shared/form/prompt.component.css
```diff
@@ -0,0 +1,82 @@
+/*
```
sandikbarr: This is just a copy of core modal css. Plan to merge functionality and remove.

src/assets/prosemirror/menu.css
```diff
@@ -0,0 +1,147 @@
+.ProseMirror-textblock-dropdown {
```
sandikbarr: These vendor styles really aren't used outside the component. Maybe they belong there instead?

src/app/shared/wysiwyg/wysiwig.component.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  cit('adds image option once images have been loaded', (fix, el, comp) => {
+    comp.name = 'description';
+    comp.model = {description: ''};
+    spyOn(comp, 'insertImageItem');
+    comp.images = [{filename: 'TestImage.jpg'}];
+    fix.detectChanges();
+    expect(comp.insertImageItem).toHaveBeenCalled();
+  });
+
+  cit('doesn\'t add empty links', (fix, el, comp) => {
```
sandikbarr: I couldn't to get a test to work adding a non empty link, so this isn't exactly a lie. I didn't want to stub out prosemirror because it is a lot, so testing was kind of a disappointment.

src/typings.d.ts
```diff
@@ -63,3 +63,11 @@ declare namespace AV {
     stop();
   }
 }
+
```
sandikbarr: Paranoid that I was missing other errors that appeared when running tests, I declared these to get the Cannot find module errors to go away. However, then I just started seeing a Can't find name error instead. We're gonna need those typings.

src/app/shared/wysiwyg/prompt.component.ts
```diff
@@ -0,0 +1,36 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'publish-prompt',
+  template: `
+  <div *ngIf="visible" class="overlay"></div>
+  <div class="modal" tabindex="-1"
+       [ngStyle]="{'display': visible ? 'flex' : 'none', 'opacity': visibleAnimate ? 1 : 0}">
```
cavis: Does `[ngStyle]` work any differently than `[style.display]="visible ? 'flex' : 'none'"]`?

src/app/shared/wysiwyg/wysiwig.component.spec.ts
cavis: Gotcha.  Integration tests will be nice... someday...

src/assets/prosemirror/menu.css
cavis: Yeah, I think these can be dropped into the component.  It's a more obvious place to go edit them.

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -0,0 +1,392 @@
...
+      </div>
+      <div class="modal-footer">
+        <button (click)="createLink()">Okay</button>
+        <button (click)="prompt.hide()">Cancel</button>
+      </div>
+    </publish-prompt>
+  `
+})
+
+export class WysiwygComponent implements OnInit, OnChanges, OnDestroy {
```
cavis: Holy state management!  That's a whole ton of stuff to handle for prosemirror!

To trim down this class a bit ... I'm wondering if some of the more image-y / state-management-y / keymap-y stuff could be moved to external classes/modules.  Something more "mixin" like, but could be accomplished with basic classes that you pass a few parameters to.  (Since we don't actually have mixins).

cavis:
So my overall CSS nitpicks (feel free to ignore until we actually iterate on this a bit):

- ProseMirrorContent is shorter than overall component (this i think should be fixed ... probably just move the min-height to ProseMirrorContent)
- Content has an outline when focused - our other fields show focus by changing the text color
- Font color/style/size should match other form fields
- Content needs a padding/line-height to match other fields
- Component overall border should match other fields
- Add red/orange outlines to overall component to reflect changed/invalid state of the field
- Validation messages not displayed (though maybe a non-issue, if we don't actually want to validate the length of these description fields)

src/assets/prosemirror/menu.css
sandikbarr: This turned into a big scoping problem and needing to add deep selectors on everything. I'm gonna leave it as global css.

cavis: Ah, right... forgot about that.  So you'd have to add `>>>` to absolutely everything.  Which makes me cringe a little every time I use them (kind of like adding `!important` to something).  Global it is then!

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -0,0 +1,392 @@
...
+import { PromptComponent } from './prompt.component';
+
+@Component({
+  selector: 'publish-wysiwyg',
+  template: `
+    <div #contentEditable></div>
+    <publish-prompt #prompt>
+      <h1 class="modal-header">Link to</h1>
+      <div class="modal-body">
+        <label>URL<span class="error" [ngStyle]="{display: isURLInvalid() ? 'inline' : 'none'}">*</span></label>
```
sandikbarr: I must be doing something bad here. Changing this to [style.display] or *ngIf results in an "Expression has changed after it was checked" error. I kinda thought maybe I shouldn't be reading native element classNames, but the `<p class="error" [style.display]="isURLInvalid() ? 'block' : 'none'">URL is required</p>` doesn't seem to have a problem with it.

cavis: I'm guessing it's because the `#url` is evaluated after this in the change detection cycle ... and since you're looking at the nativeElement, it could change after this was evaluated.  Perhaps a difference in how angular does `ngStyle` vs `style.display`?

Rather than going to the native element, you could just regex-check the string value.  Probably safer than directly looking at the DOM, as far as change detection goes.

Also, how is this currently determining if the URL is valid or not?  It's a plain `<input>` element, tied to a string `linkURL` - so I'm not sure why it would get marked as `ng-invalid` in the first place.

sandikbarr: It's getting ng-invalid when empty because of the required attribute. We could use something like validatejs to check the URL https://validatejs.org/#validators-url

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -0,0 +1,399 @@
...
+    }
+    return this.cmdItem(toggleMark(markType), passedOptions);
+  }
+
+  wrapListItem(nodeType, options) {
+    return this.cmdItem(wrapInList(nodeType, options.attrs), options);
+  }
+
+  isURLInvalid() {
+    return this.url.invalid && this.url.dirty;
```
sandikbarr: Lesson learned on checking the nativeElement.className, much safer to use the getters provided instead.

sandikbarr: This still isn't checking for anything more than non empty string (with the required attribute.) I've used https://validatejs.org/#validators-url for this before, might be an option but it would also limit users entering relative URLs or #hash links, not sure if those would be useful.

sandikbarr: One thing I would probably try if I put in more work on the linking specifically is rather than just toggling off the mark/link when clicking link on already linked text, it would read the link and fill in the URL and title in the prompt. Seems like this would be more useful since there's no way for the user to verify a link after they have entered it, but it also seems like it could be painful to implement.

cavis: Ah, I didn't even know you could do that. 👍

cavis: I say leave it for now.  Easy PR to do later on.  And we need to start stockpiling some easy things to work on!

cavis:
Wow - all of my CSS nitpicks resolved!  Looks great.

Found 1 more bug: clicking "Discard" gets rid of the orange changed outline, but doesn't update the content in the editor.

sandikbarr:
ProseMirror code moved into a class that doesn't know about much about the angular component and component doesn't know much about ProseMirror.

- [ ] New bug though: "Expression has changed after it was checked" really has it out for me today. This error comes up when you delete the story image. Wysiwyg component is in the stack trace, so definitely related.
- [ ] Discard bug is also important to address.

cavis:
Hmm... can you document exactly how you produce that "Expression has changed after it was checked" error?  I'm having trouble getting that when editing an existing story, with various combinations of adding/removing images and saving/discarding.

I do see a "expression has changed after it was checked" error every so often, but coming from the `StoryHeroComponent`, not the editor.  Going to see if I can reproduce/fix that in #88.

sandikbarr:
The StoryHeroComponent Expression has changed error is the one I'm getting now. I do think there was a different stack trace with Wysiwyg component at one point, but also I might be delusional. I did check the master branch and the StoryHeroComponent error is currently happening when you edit a story without an image, add an image, then remove the image.

I will look into using `ngDoCheck` for the change detection. I'm having issues with the Discard feature where change detection is not seeing model changes in the ngOnChanges changes parameter.

Issue #85, opened 2016-11-04
Recover from auth expiration


cavis:
If you sit idle for long enough, and then navigate without refreshing the page, you'll see this error:

```
Got 401 from GET https://cms-staging.prx.tech/api/v1/authorization
```

Should gracefully recover from this, by re-requesting a new auth token via the iframe.

PR #104, opened 2016-11-30
Recover from token expiration

cavis:
See #85.

This requires PRX/cms.prx.org#190 and PRX/id.prx.org#34 to run CMS and ID locally.

- [x] Rework how tokens are tied to the HTTP `halremote` layer, in order to refresh an expired token and seamlessly retry the request.
- [x] Fixed some bugs in how the `halremote` caching/memoization worked.
- [x] More tests!

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -13,7 +13,13 @@ export class AppComponent {
   constructor(authService: AuthService, cmsService: CmsService) {
     authService.token.subscribe((token) => {
       if (token) {
-        cmsService.setToken(token);
+        let refresher: any;
+        cmsService.setToken(token, cb => {
+          if (!refresher) {
+            refresher = authService.refreshToken();
+          }
+          return refresher;
+        });
       }
       this.loggedIn = token ? true : false;
     });
--- a/src/app/core/cms/halremote.cache.ts
+++ b/src/app/core/cms/halremote.cache.ts
@@ -47,22 +47,28 @@ export class HalRemoteCache {
     if (Env.CMS_TTL === 0) {
       return valObservable;
     }
-    this.OBSERVABLES[key] = valObservable.share();
-    this.OBSERVABLES[key].subscribe(
-      (val: any) => {
-        if (Env.CMS_USE_LOCALSTORAGE && window && window.localStorage) {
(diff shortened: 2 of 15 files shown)


src/app/app.component.ts
```diff
@@ -13,7 +13,13 @@ export class AppComponent {
   constructor(authService: AuthService, cmsService: CmsService) {
     authService.token.subscribe((token) => {
       if (token) {
-        cmsService.setToken(token);
+        let refresher: any;
+        cmsService.setToken(token, cb => {
```
cavis: The CMS and Auth services were purposefully decoupled.  But I needed some way for CMS to request that Auth "go get a new token".

This seemed like the best place to inject that callback function into CMS ... since this was already telling CMS about any new auth tokens.

I wish this was cleaner, but not sure I have a better idea for this bi-directional communication.

src/app/core/cms/halremote.cache.ts
```diff
@@ -47,22 +47,28 @@ export class HalRemoteCache {
...
-          this.EXPIRES[key] = new Date().getTime();
-        }
-        setTimeout(() => { delete this.OBSERVABLES[key]; }, 0);
-      },
-      (err: any) => {
-        setTimeout(() => { this.del(key); }, 0);
+
+    // monitor observable for value
+    let gotValue = false;
+    this.OBSERVABLES[key] = valObservable.share().map(val => {
```
cavis: Spent a long time trying to refactor this method, to work with the auth-token changing.  The keys were (1) don't subscribe, just `map`, and (2) call `share()` before anything else.  Otherwise, I got strange results where it cached the result of the token-lookup-observable, but ran the HTTP request every single time.  (Even thinking about this enough to write this comment makes my head hurt).

src/app/core/cms/halremote.ts
```diff
@@ -12,7 +12,8 @@ export class HalRemote {
   constructor(
     private http: Http,
     private host: string,
-    private token: string
+    private token?: Observable<string>,
```
cavis: This is the key here: the `token` is now an observable rather than static string.  As it might change down the road, if the old token expired.

PR #105, opened 2016-12-05
Image removal change detection

cavis:
It was a "Really Bad Idea of Ryan's" (™) to alter the `images` array during the call to `get noImages()`.  Because it occurs during change detection.  Which really shouldn't change things.  Sad.

And more generally, a call to `model.changed()` should return false if the model is deleted AND new.


cavis:
And as mentioned in #103 - you can produce this bug by adding-then-removing an image to any story that doesn't have an image.

cavis:
Update: I think I added another bug.  Localstorage seems to be holding onto the upload after deletion.

gcampo88:
👍

Issue #97, opened 2016-11-04
Audio delete prompt


cavis:
Display a prompt when deleting an audio file.

...OR...

Provide some sort of an undo button.  (Deletes aren't propagated until you hit save anyways).

PR #107, opened 2016-12-05
Prompt before deleting audio files

cavis:
See #97.

Not sure of the best wording here, so I made it up.

There are actually a couple different scenarios around deleting an audio file:

1. New unsaved audio - we really do need to prompt for this, as clicking the X will remove all traces forever.
2. Saved audio - also prompting for this, but not quite as catastrophic, since your deletion won't be sent to the server until you click "Save".  (And you can always "Discard" first).
3. In-progress upload - not prompting for this.  Just cancel the upload.
4. Failed upload - not prompting for this, as there's nothing to save.

### DIFF
--- a/src/app/upload/shared/audio-cancel.directive.spec.ts
+++ b/src/app/upload/shared/audio-cancel.directive.spec.ts
@@ -1,6 +1,6 @@
 import { cit, create, direct, provide, By } from '../../../testing';
 import { Component } from '@angular/core';
-import { ModalService } from '../../core'
+import { ModalService } from '../../core';
 import { AudioCancelDirective } from './audio-cancel.directive';
 
 @Component({
--- a/src/app/upload/shared/audio-cancel.directive.spec.ts
+++ b/src/app/upload/shared/audio-cancel.directive.spec.ts
@@ -1,5 +1,6 @@
-import { cit, create, direct, By } from '../../../testing';
+import { cit, create, direct, provide, By } from '../../../testing';
 import { Component } from '@angular/core';
+import { ModalService } from '../../core'
 import { AudioCancelDirective } from './audio-cancel.directive';
 
 @Component({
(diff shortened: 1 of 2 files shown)


src/app/upload/shared/audio-cancel.directive.ts
```diff
@@ -12,7 +13,21 @@ export class AudioCancelDirective {
...
+  constructor(private modal: ModalService) {}
+
   @HostListener('click') onClick() {
+    if (this.publishAudioCancel.isUploading) {
+      this.cancelAndDestroy();
+    } else {
+      this.modal.prompt(
+        'Really delete audio file?',
+        'This operation cannot be undone',
+        (okay: boolean) => okay && this.cancelAndDestroy()
```
gcampo88: might vote action instead of operation

src/app/upload/shared/audio-cancel.directive.spec.ts
```diff
@@ -1,5 +1,6 @@
-import { cit, create, direct, By } from '../../../testing';
+import { cit, create, direct, provide, By } from '../../../testing';
 import { Component } from '@angular/core';
+import { ModalService } from '../../core'
```
gcampo88: semicolon!

src/app/upload/shared/audio-cancel.directive.spec.ts
```diff
@@ -10,20 +11,44 @@ class MiniComponent {
...
   delay = 10;
 }
+const getDirective = (el): AudioCancelDirective => {
+  let cancelEl = el.query(By.directive(AudioCancelDirective));
+  expect(cancelEl).toBeDefined();
+  return cancelEl.injector.get(AudioCancelDirective);
+};
 
-describe('AudioCancelDirective', () => {
+fdescribe('AudioCancelDirective', () => {
```
gcampo88: need to nix `fdescribe` here before merging

src/app/upload/shared/audio-cancel.directive.spec.ts
cavis: Whoops!  I should pay more attention to tslint complaints.

src/app/upload/shared/audio-cancel.directive.spec.ts
cavis: What, you mean I have to get all the tests to pass, not just this one?

PR #135, opened 2016-12-09
Fix evaporate 403s

cavis:
Publish encounters 403s after uploading a file, canceling mid-upload, then uploading the same file again.  The file must be sufficiently large that Evaporate splits it into multiple parts.

It seems it was trying to reuse the old upload parts because the file was identical, even though the S3 upload path (which includes a Publish-generated UUID) was different.  Setting `onlyRetryForSameFileName = true` forced Evaporate to only reuse the parts if the entire S3 key matches.  (Which it won't because of the guid).

Also see the [Evaporate 1.x docs](https://github.com/TTLabs/EvaporateJS/tree/7e6e65ad49ba3260e841d3de86c2984bc52af9c3#api), and [this github issue](https://github.com/TTLabs/EvaporateJS/issues/210).

This does not change the intermittent GET 403s seen in the console, which seem to be the result of Evaporate trying to clean up a canceled upload.  (The DELETEs all succeed, and then there's a final GET for some reason).  But since this error is presumably caught somewhere, and it's just the non-200 XHR showing up in the console, I'm not really worried about it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "connect-gzip-static": "^1.0.0",
     "core-js": "^2.4.1",
     "dotenv": "^2.0.0",
-    "evaporate": "^1.5.8",
+    "evaporate": "^1.6.3",
     "express": "^4.14.0",
     "express-http-proxy": "^0.10.1",
     "morgan": "^1.7.0",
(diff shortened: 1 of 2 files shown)


cavis:
Sending to @kookster to verify this is sane.  Or at least less-insane than usual.

Issue #127, opened 2016-12-08
Wysiwyg discard after save


cavis:
This is sort of an edge case.  Discarding changes on the wysiwyg editor always reverts to the initial (when you first navigated to that page) state.  Steps to produce...

1. Change a story description
2. Click save
3. Change it again
4. Click discard
5. Reverts to the initial state from before (1)

PR #138, opened 2016-12-12
Wysiwyg discard after save

sandikbarr:
This addresses bug #127 by adding change detection on the primitive string value as well as binding to model changed.

### DIFF
--- a/src/app/shared/wysiwyg/wysiwyg.component.ts
+++ b/src/app/shared/wysiwyg/wysiwyg.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit, OnChanges, OnDestroy, ElementRef, ViewChild, SimpleChanges } from '@angular/core';
+import { Component, Input, OnInit, OnChanges, SimpleChanges, OnDestroy, ElementRef, ViewChild } from '@angular/core';
 import { NgModel } from '@angular/forms';
 import { BaseModel } from '../model/base.model';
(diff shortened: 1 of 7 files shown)


src/app/story/directives/basic.component.ts
```diff
@@ -24,7 +24,8 @@ import { StoryModel, TabService } from '../../shared';
         <label for="description">Description</label>
       </h3>
       <p class="hint">A full description of your piece.</p>
-      <publish-wysiwyg *ngIf="story" [model]="story" name="description" [images]="story.images"></publish-wysiwyg>
+      <publish-wysiwyg *ngIf="story" [model]="story" name="description" [content]="story.description" [images]="story.images" 
+        [changed]="descriptionChanged"></publish-wysiwyg>
```
cavis: Could also have used just `story.changed('description')` here (since there's already an ng-if-story) ... but this is equivalent.

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -31,8 +31,10 @@ import { PromptComponent } from './prompt.component';
 export class WysiwygComponent implements OnInit, OnChanges, OnDestroy {
   @Input() model: BaseModel;
   @Input() name: string;
+  @Input() content: string;
```
cavis: Ah, so it turned out you needed that string as an `@Input` here.  Makes sense.

PR #139, opened 2016-12-13
Fix wysiwyg Cancel add link expression changed error

sandikbarr:
- [ ] fixes the Expression changed after checked error when the user opens the link prompt, doesn't fill in the fields, and clicks Cancel
- [ ] warns the user when they have clicked the link button if no text is selected

### DIFF
--- a/src/app/shared/wysiwyg/prompt.component.ts
+++ b/src/app/shared/wysiwyg/prompt.component.ts
@@ -27,6 +27,6 @@ export class PromptComponent {
   }
 
   public hide(): void {
-    setTimeout(() => this.visible = false, 300);
+    this.visible = false;
   }
 }
--- a/src/app/shared/wysiwyg/wysiwyg.component.ts
+++ b/src/app/shared/wysiwyg/wysiwyg.component.ts
@@ -10,7 +10,8 @@ import { PromptComponent } from './prompt.component';
   template: `
     <div #contentEditable [class.changed]="changed" [class.invalid]="invalid"></div>
     <p *ngIf="invalid" class="error">{{invalid | capitalize}}</p>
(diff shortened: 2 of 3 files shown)


src/app/shared/wysiwyg/prompt.component.ts
```diff
@@ -27,6 +27,6 @@ export class PromptComponent {
   }
 
   public hide(): void {
-    setTimeout(() => this.visible = false, 300);
```
sandikbarr: This right here was the source of the Expression changed after checked. I was thrown off by the stack trace because it didn't show the Prompt component, it showed the Wysiwyg component. That's what I get for using an example from Stack Overflow.

gcampo88: wow. good sleuthing!

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -24,6 +25,15 @@ import { PromptComponent } from './prompt.component';
         <button (click)="prompt.hide()">Cancel</button>
       </div>
     </publish-prompt>
+    <publish-prompt #prompt *ngIf="editor?.isSelectionEmpty()">
+      <h1 class="modal-header">Warning</h1>
+      <div class="modal-body" *ngIf="editor?.isSelectionEmpty()">
```
gcampo88: any reason for the duplicated `*ngIf="editor?.isSelectionEmpty()"` here + on line 28?

sandikbarr: oops, thought I moved it, turns out I duplicated it. Fixing...

PR #143, opened 2016-12-14
fix missing draft info on dashboard

### LINKED ISSUE #90: Dashboard icons missing some draft data
- [x] Audio durations on the dashboard not including unsaved audio
- [x] Make sure unsaved images are shown on the dashboard
- [x] Make sure you get a dashboard draft, even if you haven't added a title yet (currently not showing as draft on dashboard if only audio, image, and/or podcast distribution is added, but is saved in local storage)

sandikbarr:
Fixes #90 dashboard draft info.
 - [ ] Audio durations on the dashboard not including unsaved audio --> should compute duration for files !isDestroy(), including those that are only in local storage and not yet persisted to CMS
 - [ ] Make sure unsaved images are shown on the dashboard --> If only an image was set, the dashboard tile was showing as an add new story plus-sign button
 - [ ] Make sure you get a dashboard draft, even if you haven't added a title yet --> If only image, audio, and or clean/explicit was set, the dashboard tile was showing as an add new story plus-sign button

### DIFF
--- a/src/app/home/directives/home-story.component.ts
+++ b/src/app/home/directives/home-story.component.ts
@@ -74,7 +74,7 @@ export class HomeStoryComponent implements OnInit {
     let duration = 0;
     if (this.story.versions && this.story.versions.length > 0) {
       if (this.story.versions[0].files.length > 0) {
-        duration = this.story.versions[0].files.map((audio) => {
+        duration = this.story.versions[0].files.filter((audio) => !audio.isDestroy).map((audio) => {
           return audio['duration'] || 0;
         }).reduce((prevDuration, currDuration) => {
           return prevDuration + currDuration;
--- a/src/app/home/directives/home-story.component.ts
+++ b/src/app/home/directives/home-story.component.ts
@@ -1,6 +1,4 @@
 import { Component, Input, OnInit } from '@angular/core';
-
-import { HalDoc } from '../../core';
 import { StoryModel } from '../../shared';
 
 @Component({
@@ -15,7 +13,7 @@ import { StoryModel } from '../../shared';
     </template>
     <template [ngIf]="!isPlusSign">
       <a [routerLink]="editLink">
-        <publish-image [src]="storyImage" [imageDoc]="storyImageDoc"></publish-image>
+        <publish-image [src]="storyImage" ></publish-image>
       </a>
       <h2>
         <a *ngIf="storyTitle" [routerLink]="editLink">{{storyTitle}}</a>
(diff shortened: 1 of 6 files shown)


src/app/home/directives/home-story.component.ts
```diff
@@ -95,4 +65,30 @@ export class HomeStoryComponent implements OnInit {
     }
   }
 
+  get isPlusSign(): boolean {
+    return this.story.isNew && !this.story.isStored()
```
sandikbarr: sadly, I have realized this is incorrect :(

cavis: maybe `this.story.isNew && !this.story.changed()` ?

cavis: i was purposefully avoiding calling `changed()` earlier, as i didn't want to load-related.  But now that we have to... may as well.

sandikbarr: That works. Thanks. Now I just need to figure out what to do if they set Explicit/Clean and nothing else, or maybe why Explicit/Clean isn't registering as changed...

src/app/shared/model/audio-version.model.ts
```diff
@@ -127,7 +127,7 @@ export class AudioVersionModel extends BaseModel {
   }
 
   changed(field?: string | string[], includeRelations = true): boolean {
-    if (this.isNew && this.files.length === 0) {
+    if (this.isNew && this.files.length === 0 && this.explicit === undefined) {
```
sandikbarr: This also has the effect of prompting the user on story edit that they have unsaved changes if they have set Clean/Explicit and nothing else, which seems correct to me.

cavis: Hmm... having to manually check for explicit here seems like a symptom of something else.

Maybe try giving it a default value of `""` (empty string) at the top of this file?  Could also mess around with the decode/encode functions, to allow a blank value rather than defaulting to 'Clean'.

Or maybe if clean really is the default, the top should be `public explicit = 'Clean'`?

sandikbarr: On the question of what the default should be, do we support not including the tag `<itunes:explicit>` if explicit is not set or is empty string? Or do we always include the tag?

gcampo88:
this looks good to me! @sandikbarr are you planning to rework the `explicit` options as default `''`, vs. `explicit`/`clean` in this PR or should that be its own small PR?

sandikbarr:
Had to get tests working again, but this is now updated with an empty string default for explicit and a falsey check on explicit in `changed()`

Issue #136, opened 2016-12-09
Should validate image file type


gcampo88:
Should constrain image file types to jpg and png both by limiting the file selection to image mime types, and show a file type-specific error message when unacceptable file type is chosen

sandikbarr:
Now that I've looked up the iTunes image requirements multiple times and it's staring me in the face, it bugs me a little bit that the image size, i.e. 1400x1400 doesn't have any spaces in it. If it doesn't bug anyone else, it's probably fine. But if you want to adjust it while working on the mime type validation, that's definitely cool with me. 😃

PR #148, opened 2016-12-16
Fix/validate image type

gcampo88:
- [X] validates upload file type, restricts image upload to jpeg/png, and shows file-type specific error message to user on attempted upload of invalid file (#136)
- [X] series image now links to series edit page instead of "#"

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "express-http-proxy": "^0.10.1",
     "morgan": "^1.7.0",
     "newrelic": "1.34.0",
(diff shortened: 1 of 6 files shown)


gcampo88:
Hmm. Currently the user is able to select any kind of file type, but will get the error message (& upload will not begin) if invalid file type. Should we also throw a `accept="image/*"` on the input field to keep them from even selecting non-img files in the first place? Audio input field has `accept="audio/mpeg" `.

cavis:
I say yes to both.  Belt AND suspenders!

Issue #125, opened 2016-12-07
Series podcast tab and fields


cavis:
Add the fields to be edited for distributing a series as a podcast, save them to the feeder podcast api via the link in the series resource.

- [x] vanity path
- [x] itunes categories
- [x] itunes explicit
- ~~[ ] feedburner url~~, 
- [x] see feeder rss url
- [x] new_feed_url for moving feed, 
- [x] homepage link

PR #149, opened 2016-12-16
Series podcast distribution tab

cavis:
See #125.

Add a podcast tab to the series page, which allows creating/saving podcast data directly to feeder.

For simplicity, you have to have a saved-series before you're allowed to create a podcast distribution for it.  And once you have a podcast distribution, you can't get rid of it.  It'll follow you to your grave.

Note that after creating the podcast, you won't be able to view either the published or preview feeds.  Published-feed is an S3 bucket that doesn't seem to exist yet.  And preview is blocked by PRX/feeder.prx.org#155 for the moment.  But you can still see the data in the feeder API.

I've also noticed that changes to normal series fields (title, description, etc) aren't propagating to feeder after the initial creation.  Probably some issue with SNS messaging.  But this at least makes it easier to debug.


src/app/shared/model/distribution.model.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  let podcastUrl = 'http://some.where/your/podcast/1234';
+  let series = cms.mock('prx:series', {id: 'series1'});
+  let fooDist = series.mock('prx:distributions', {id: 'dist1', kind: 'foo'});
+  let podDist = series.mock('prx:distributions', {id: 'dist2', kind: 'podcast', url: podcastUrl});
+  let podcast = podDist.mock(podcastUrl, {id: 'pod1'});
+
+  beforeEach(() => window.localStorage.clear());
+
+  it('loads an no podcast by default', () => {
```
gcampo88: "an no" = "a new" ?

cavis: whoops!

src/app/series/directives/series-podcast.component.html
```diff
@@ -0,0 +1,65 @@
...
+        <publish-fancy-field label="Preview Feed" small=1>
+          <input type="text" readonly [ngModel]="podcast?.previewUrl" name="previewUrl"/>
+          <a class="button" target="_blank" [href]="podcast?.previewUrl">Open Link</a>
+        </publish-fancy-field>
+      </div>
+    </publish-fancy-field>
+
+    <publish-fancy-field label="iTunes Categories" required=1>
+      <div class="fancy-hint">Some description of what these are here and
+        maybe also a link to the itunes docs.</div>
```
cavis: Wasn't really sure what to use for placeholder text in here.  Every one of these fields could link to the itunes docs, but that seems a bit redundant.

src/app/series/directives/series-podcast.component.html
```diff
@@ -0,0 +1,65 @@
...
+        <publish-fancy-field *ngIf="subCategories.length" label="Sub-Category"
+          [model]="podcast" name="subCategory" [select]="subCategories" small=1>
+        </publish-fancy-field>
+      </div>
+    </publish-fancy-field>
+
+    <publish-fancy-field label="Explicit" required=1 [model]="podcast" name="explicit" [select]="explicitOpts">
+      <div class="fancy-hint">Do any of your podcast episodes contain explicit material?</div>
+    </publish-fancy-field>
+
```
cavis: Yeah, we should open another ticket for that.  It's a bit complicated since the value of this field is currently stored in feeder, not CMS.  But worth looking into.

src/app/shared/form/copy-input.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+  private copiedText = 'Copied';
+
+  constructor(private el: ElementRef) {}
+
+  @HostListener('mouseover') onMouseOver(): void {
+    this.el.nativeElement.innerHTML = this.startText;
+  }
+
+  @HostListener('click') onClick(): void {
+    if (this.publishCopyInput && this.publishCopyInput.select) {
```
sandikbarr: From what I recall, with this same code in play.prx.org TypeScript thought that publishCopyInput was of type HTMLAnchorElement (even though we know it's a HTMLInputElement) and was throwing console errors because HTMLAnchorElement did not have a select() function. Not sure if it is complaining about this cuz I can't run it right now...

cavis: That does sound familiar.  I didn't notice any compilation errors on this, though.  Maybe because I'm passing it in via this directive's `@Input`... typescript doesn't have any choice but to trust me on what type it is?

sandikbarr: saved by the type declaration!

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -0,0 +1,119 @@
...
+  encode(): {} {
+    let data = <any> {};
+
+    // unset things with nulls instead of blank strings
+    data.explicit = this.explicit || null;
+    if (data.explicit) {
+      data.explicit = data.explicit.toLowerCase();
+    }
+    data.link = this.link || null;
+    this.newFeedUrl = this.newFeedUrl || null;
```
gcampo88: typo! should be `data.newFeedUrl = this.newFeedUrl || null`, no? (I was wondering why my http://puppies.com wasn't persisting!)

cavis: Good catch!

PR #152, opened 2016-12-19
edit links in prosemirror wysiwyg editor

sandikbarr:


src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -106,7 +108,10 @@ export class WysiwygComponent implements OnInit, OnChanges, OnDestroy {
     return this.url && this.url.invalid && this.url.dirty;
   }
 
-  promptForLink() {
+  promptForLink(url?: string, title?: string) {
+    this.linkURL = url;
+    this.linkTitle = title;
+    this.chgRef.detectChanges();
```
sandikbarr: seems to have introduced an Expression Changed After Checked Error if the user hits the Enter button instead of clicking Okay on the prompt

src/app/shared/wysiwyg/prompt.component.ts
```diff
@@ -18,15 +18,4 @@ import { Component } from '@angular/core';
   `,
   styleUrls: ['prompt.component.css']
 })
-export class PromptComponent {
```
sandikbarr: I made a terrible error in the way I managed visible state in this component. I was calling show and hide from the #prompt template variable in the parent component. This bypasses change detection. 

I could have passed visible in as an input, but I already had an ngIf in the parent, so I just used that. Seems like it makes no sense to have this as a separate component now, so I'm going to get rid of it.

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -1,39 +1,40 @@
...
 import { ProseMirrorMarkdownEditor, ProseMirrorImage } from './prosemirror.markdown.editor';
-import { PromptComponent } from './prompt.component';
 
 @Component({
   selector: 'publish-wysiwyg',
   template: `
     <div #contentEditable [class.changed]="changed" [class.invalid]="invalid"></div>
     <p *ngIf="invalid" class="error">{{invalid | capitalize}}</p>
     
-    <publish-prompt #prompt *ngIf="!editor?.isSelectionEmpty()">
```
sandikbarr: I made a terrible error in the way I managed visible state in the prompt component. I was calling show and hide from the #prompt template variable in this parent component. This bypasses change detection.

I could have passed visible in as an input to prompt component, but it stopped making sense to keep it as a separate component. The only benefit seems like a separate place to keep modal CSS, but it was using deep selectors, so... not ideal.

So basically, prompt.component can die in a fire (and I can learn to use Angular properly.)

gcampo88:
looking good! a few things I've noticed so far -- let me know if you'd like to address now / leave for later / whether they're desired behavior:
- when cursor is at the beginning of a hyperlinked word, prosemirror doesn't seem to know it (if I click the hyperlink button then, I get the warning asking me to select text); but it does know about the selection if the cursor is at the end of the hyperlinked word. asymmetry feels slightly odd. 
- If I make a hyperlink in my description text and then continue adding text after the hyperlink (as opposed to writing all of my text and then going back in to select phrases to hyperlink), the added text becomes part of its preceding hyperlink instead of just plain text. 
- no current way to remove a hyperlink once created?
- clicking Enter key while creating or editing a hyperlink doesn't save + exit prompt -- instead, adds newlines to Description field behind prompt. maybe input fields could autofocus when prompt opens?

sandikbarr:
- [ ] I noticed this about being at the beginning of a link too. I might be able to trick prosemirror's mark detection into thinking it's on the mark when at the beginning of a link, but it's gonna be fussy. I'll give it a shot and report back.
* The inclusion of new text into the end of an existing mark (link) is fully controlled by prosemirror. It's not a good solution at all, but if you break the mark by typing on a new line, you can remove the line break and get to where you were going.
* Yeah, this removed the toggle link mark off functionality. The links now can be removed by removing the linked text. Rather than just handling toggle off for links, it seems like it would be possible to add another button to the toolbar that would clear any kind of mark/formatting. If we want to go that route, I'd prefer to make it a separate issue/PR.
- [ ] Yikes. Definitely needs to focus on the url when opening prompt. Good call!

PR #156, opened 2016-12-20
Multiple series images

cavis:
Allows adding 2 images for the series.  One (the current "Cover Image") will be set to the big `<itunes:image>` in the feed.  The other (the small "Thumbnail") will be set to the tiny `<image>` in the feed.

Required some changes to how the SeriesModel tracks new ImageModels.  The SeriesModel (and StoryModel for that matter) now track a settable property for `uploadUUIDs`, much the same way the AudioVersionModel tracked AudioFileModels.

Abstracted this "has many uploads" concept into a mixin.  A bit ugly, but it works.


src/app/core/cms/haldoc.ts
```diff
@@ -108,11 +108,11 @@ export class HalDoc {
     }
   }
 
-  has(rel: string): boolean {
+  has(rel: string, mustBeList = false): boolean {
```
cavis: There's still one instance in CMS (the audio-version's `_embedded` audio-files) where the contents aren't behind a `prx:items` array.  And I needed the ability to check for that, for now.

src/app/shared/model/audio-version.model.ts
```diff
@@ -3,19 +3,18 @@ import { HalDoc, Upload } from '../../core';
...
 
   public id: number;
-  public uploads: string = '';
   public label: string;
   public explicit: string = '';
-  public files: AudioFileModel[];
 
   // save in-progress uploads to localstorage
-  SETABLE = ['uploads', 'label', 'explicit'];
+  SETABLE = ['label', 'explicit', 'hasUploadMap'];
```
cavis: It's kind of a bummer that I have to put `hasUploadMap` here.  And re-define the other 3 methods from that mixin below.  But unfortunately that's the only way to make typescript play nice with them.

src/app/shared/model/image.model.ts
```diff
@@ -36,12 +34,10 @@ export class ImageModel extends UploadableModel {
   key() {
     if (this.doc) {
       return `prx.image.${this.doc.profileSubtype}.${this.doc.id}`;
-    } else if (this.parent) {
-      return `prx.image.new.${this.parent.profileType}.${this.parent.id}`;
-    } else if (this.grandparent) {
-      return `prx.image.draft.${this.grandparent.profileType}.${this.grandparent.id}`;
+    } else if (this.uuid) {
+      return `prx.image.new.${this.uuid}`;
```
cavis: Much easier to use a uuid in the key, instead of relying on the parent/grandparent models.  Just requires that the parent model track which new-upload-uuids it has.

src/app/shared/model/story.model.ts
```diff
@@ -134,9 +135,20 @@ export class StoryModel extends BaseModel {
     }
   }
 
-  get unsavedImage(): ImageModel {
-    let img = new ImageModel(this.parent, this.doc, null);
-    return img.isStored() && !img.isDestroy ? img : null;
+  addImage(upload: Upload): ImageModel {
+    let image = new ImageModel(this.doc, upload);
+    this.images.push(image);
```
sandikbarr: I'm finding that change detection isn't working for the wysiwyg editor insert image menu. Can addImage and removeImage set this.images to new arrays instead (both for story and series)?
`this.images = [...this.images, image];`
`this.images = [...this.images.splice(this.images.indexOf(image), 1)];`
or `slice` or whatever you like 😃

cavis: Ah, good call.  I saw a change detection bug coming from there at one point, but assumed it was a fluke!

Will try making a new array.  Which is probably a good policy anytime we're adding/removing a model from the array.

src/app/shared/wysiwyg/wysiwyg.component.ts
```diff
@@ -70,8 +70,7 @@ export class WysiwygComponent implements OnInit, OnChanges, OnDestroy {
 
   ngOnChanges(changes: SimpleChanges) {
     if (this.editor) {
-      let imagesChanged = changes['images'] && changes['images'].currentValue.length !== changes['images'].previousValue.length;
-      if (imagesChanged) {
+      if (changes['images']) {
```
sandikbarr: This scares me just a little. I'll do some testing on it. When I built this, images seemed to always be in the changes even when there were no changes and it fires constantly, which means editor menu is being rebuilt with lots of DOM changes.

sandikbarr: Testing says 100% better change detection on images!

Issue #167, opened 2016-12-21
Series, Podcast Distribution, author info


kookster:
Show the author info that will be published for the podcast based on the account
Allow the user to override this information and set the author full name and email.

gcampo88:
based on [the example of 99pi](https://feeder.prx.org/api/v1/podcasts/23), looks like we currently only give full name for `author`, and skip the email (`owner` has both name and email): `"owner": {
    "name": "Roman Mars",
    "email": "roman@prx.org"
  },
  "author": {
    "name": "Roman Mars"
  },`

So, assume we should just allow user to override author name?

cavis:
Updated requirements: Author-name is required, Author-email is optional.  So still need to add an Email field.

(Given how this form is getting fairly long, may want to consider having these 2 fields on a single line, under a single form label... but I'll leave that up to you!)

PR #177, opened 2016-12-22
Allow user to override default author name for podcast

gcampo88:
By default, `author` will be set to account holder's name. This allows the user to override the author name that will be displayed with their podcast. (See #167)

### DIFF
--- a/src/app/shared/model/feeder-podcast.model.ts
+++ b/src/app/shared/model/feeder-podcast.model.ts
@@ -55,7 +55,7 @@ export class FeederPodcastModel extends BaseModel {
     }
     this.link = this.doc['link'] || '';
     this.newFeedUrl = this.doc['newFeedUrl'] || '';
-    if (this.doc['author']['name']) {
+    if (this.doc['author'] && this.doc['author']['name']) {
       this.authorName = this.doc['author']['name'];
     } else if (this.series.has('prx:account')) {
       this.series.follow('prx:account').subscribe(account => this.authorName = account['name']);
--- a/src/app/shared/model/feeder-podcast.model.ts
+++ b/src/app/shared/model/feeder-podcast.model.ts
@@ -55,11 +55,12 @@ export class FeederPodcastModel extends BaseModel {
     }
     this.link = this.doc['link'] || '';
     this.newFeedUrl = this.doc['newFeedUrl'] || '';
-    if (this.series.has('prx:account')) {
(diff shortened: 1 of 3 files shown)


src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -54,7 +55,12 @@ export class FeederPodcastModel extends BaseModel {
     }
     this.link = this.doc['link'] || '';
     this.newFeedUrl = this.doc['newFeedUrl'] || '';
-
+    if (this.series.has('prx:account')) {
+      this.series.follow('prx:account').subscribe(account => this.authorName = account['name']);
+    }
+    if (this.doc['author']) {
+      this.authorName = this.doc['author']['name'];
+    }
```
sandikbarr: Since that subscribe is technically asynchronous and continually subscribed, potentially could be overriding this.authorName. That's not actually what's happening because prx:account is embedded, so returns immediately, and doesn't update but I would prefer to see
```
if (this.doc['author'])
  this.authorName = this.doc['author']['name'];
else
  this.series.follow('prx:account').subscribe(account => this.authorName = account['name']);
```
I'm also thinking we've got to have that account info in the app already and shouldn't have to do that follow here. We can't inject the CmsService into this model like we can components. I think if this.series were a SeriesModel its parent should be the account. Hmm... we may need @cavis's advice on this.

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -54,7 +55,12 @@ export class FeederPodcastModel extends BaseModel {
     }
     this.link = this.doc['link'] || '';
     this.newFeedUrl = this.doc['newFeedUrl'] || '';
-
+    if (this.series.has('prx:account')) {
+      this.series.follow('prx:account').subscribe(account => this.authorName = account['name']);
+    }
+    if (this.doc['author']) {
+      this.authorName = this.doc['author']['name'];
```
sandikbarr: It could be something weird about my account, but for me doc has an author but the author Object has no properties, so this.authorName is being set to undefined. (It was previously === "Sandi Barr" from the account['name'] though.)

gcampo88: Good catch!

src/app/shared/model/feeder-podcast.model.ts
gcampo88: good call. For now I'll put in if/else logic, and continue to poke around!

PR #180, opened 2016-12-23
Search updates: multiline overflow handling, series genre remove, story card series link removal

sandikbarr:
- Basically the same code, so a couple small ones are included in this PR that is primarily for multiline overflow handling.
 - I kept all the ways to indicate overflow in shared text-overflow components. The only one in use is the fade. Webkit is what we were previously using.
 - Overflow indication is only applied to description text on the search cards. Titles and tags have overflow hidden, but it is not indicated.


src/app/shared/text-overflow/text-overflow-webkit-line-clamp.component.css
```diff
@@ -0,0 +1,10 @@
+div {
```
sandikbarr: This is what we were using previously. There was some discussion of just using the webkit line clamp, so I kept it around. Considering the drawbacks on the fade and cross browser ellipses techniques, figured we might eventually want this in the style guide in cases where the other two just aren't the right choice.

gcampo88: 👍

src/app/search/directives/search-series-form.component.ts
```diff
@@ -1,7 +1,6 @@
 import { Component, Input, Output, EventEmitter, OnInit } from '@angular/core';
 import { Subject } from 'rxjs/Subject';
 import { SearchSeries } from '../search-series.model';
-import { CATEGORIES, SUBCATEGORIES } from '../../shared/model/itunes.categories';
```
gcampo88: should we remove all mention of genre from the model too? 
screen shot 2016-12-27 at 10 26 03 am

src/app/shared/text-overflow/text-overflow-ellipses.component.ts
```diff
@@ -0,0 +1,17 @@
...
+      <ng-content></ng-content>
+    </div>
+  `,
+  styleUrls: ['text-overflow-ellipses.component.css']
+})
+
+export class TextOverflowEllipsesComponent {
+  @Input() numLines: number = 3;
+  @Input() lineHeight: number = 1.2;
+  @Input() unit: string = 'em';
```
gcampo88: how come all the components have unit default to 'em' but then are set to 'px' whenever they're used?

src/app/search/directives/story-card.component.css
```diff
@@ -77,12 +77,20 @@ publish-image  {
   flex: 1 1 60px;
 }
 .audio-total > i {
-  /* TODO: this isn't the correct icon */
+  /* TODO: this isn't the correct icon -doesn't matter, this is supposed to be play count anyway #56 */
```
gcampo88: :)

src/app/shared/text-overflow/text-overflow-ellipses.component.ts
sandikbarr: Typically fonts and related items should be sized in ems for responsive layouts because smaller screen sizes should have larger fonts. But the Publish app layouts are pretty well set on a 20px line height at every screen size. These text overflow components will eventually be moved into a styleguide, and ems seemed like a more generic default for that purpose.

src/app/search/directives/search-series-form.component.ts
sandikbarr: yeah, missed it. Let me get rid of that...

src/app/shared/text-overflow/text-overflow-ellipses.component.ts
gcampo88: gotcha!

Issue #176, opened 2016-12-22
Published Feed doesn't update to include custom/vanity path at first


gcampo88:
Published Feed doesn't update to include your new custom/vanity path when you hit "save" -- only shows up once you refresh the page. could be solved by @sandikbarr 's suggestion of having input field for custom path be a part of the Published Feed input field (see #165)

sandikbarr:
This explains why the field wasn't making sense to me!

kookster:
I didn't see @sandikbarr's suggestion, but you can't edit the published url
except for the vanity path.
It is supposed to be there for display purposes, like the embed url, so you
can either use it, or submit it to feedburner.

There is a different field (I think in another ticket), related to the
embed player, to edit and manage a public url for the feed which is usually
a feedburner feed url.

Issue #170, opened 2016-12-21
Story, Podcast Distribution, remove preview feed url


kookster:
We don't need to show the preview url, I know it is useful for our own testing, but not to be exposed to users.

Issue #165, opened 2016-12-21
Series, Podcast Distribution, Vanity path changes


kookster:
- [x] Move the field up to the top with "PRX Feeds", as it affects those feed urls
- [ ] Explain in the hint how it is used instead of the id # in the PRX Feed urls
- [x] Make the text field much shorter, as this field should be only a 5-20 characters, like the feedburner feed path

sandikbarr:
For the hint, not sure how users would refer to or understand the vanity path.

Does an input like this make sense at all?
![feedburner45](https://cloud.githubusercontent.com/assets/2250413/21430509/12ec2d54-c829-11e6-914f-85b6a761ae7f.jpg)

gcampo88:
@sandikbarr I think that would make it much clearer! Also noticed just now that the Published Feed url doesn't update once you save your new Vanity/Custom Path -- only shows up in place once you refresh the page. Made a new issue for it --#176

kookster:
I read these in the wrong order, yes, that is helpful, and is the part you
can update, though it is in the middle not the end. (It is also used to
customize where the related files are saved, not just the rss xml resource.)

We also need sort of the reverse field, where you can set the prefix for
the audio files (e.g. podtrac and blubrry), I'll open a ticket.

Issue #163, opened 2016-12-21
Series, Podcast Distribution, Feed url text too light


kookster:
![image](https://cloud.githubusercontent.com/assets/46439/21404352/c561a734-c78e-11e6-9078-35896cb5f3e6.png)

The text in published and preview URLs look the same as the hint text for fields with nothing in them, use a darker text color please.

PR #181, opened 2016-12-23
Podcast Distribution custom path

sandikbarr:
#163, #165, #170, #176 

The one thing not covered here is explaining to the user that the custom path is used in the URL instead of the podcast id. I'm just still not clear on if users understand what the podcast id # is and what this hint should actually say. I think putting it in the context of the url prefix and suffix does help though (and fixing that bug where it wasn't actually showing up when set until requested from the API again.)

### DIFF
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -22,8 +22,10 @@
           <a class="button" target="_blank" rel="noopener" [href]="podcast?.publishedUrl">Open Link</a>
         </publish-fancy-field>
 
-        <publish-fancy-field label="Custom Path" textinput [model]="podcast" name="path">
+        <publish-fancy-field class="custom-path" label="Custom Path" textinput [model]="podcast" name="path" small=1>
           <div class="fancy-hint">An optional custom folder name to use in your PRX published feed url.</div>
+          <span class="prefix">{{urlStart}}</span>
+          <span class="suffix">{{urlEnd}}</span>
         </publish-fancy-field>
       </div>
     </publish-fancy-field>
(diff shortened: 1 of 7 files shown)


src/app/series/directives/series-podcast.component.html
```diff
@@ -21,9 +21,11 @@
           <button [publishCopyInput]="pubFeed">Copy</button>
           <a class="button" target="_blank" rel="noopener" [href]="podcast?.publishedUrl">Open Link</a>
         </publish-fancy-field>
-        <publish-fancy-field label="Preview Feed" small=1>
```
sandikbarr: I think it's being used for testing right now, so I was worried about removing support entirely. I was considering checking if we'd done anything with the `process.env.NODE_ENV` yet and log it out when ` !== 'production'`, but I didn't check.

src/app/series/directives/series-podcast.component.html
```diff
@@ -21,9 +21,11 @@
...
-          <input type="text" readonly [ngModel]="podcast?.previewUrl" name="previewUrl"/>
-          <a class="button" target="_blank" rel="noopener" [href]="podcast?.previewUrl">Open Link</a>
+
+        <publish-fancy-field class="custom-path" label="Custom Path" textinput [model]="podcast" name="path" small=1>
+          <div class="fancy-hint">An optional custom folder name to use in your PRX published feed url.</div>
```
sandikbarr: Weird. It's maybe something with how we're handling settable fields in models. I'll look into it.

sandikbarr: It was a combination of things. I needed to add publishedUrl to the settable fields then call set on it, and I needed to add it to the encode because the haldoc uses the data properties in the encode to set itself again after an update.

gcampo88: very elegant!

Issue #164, opened 2016-12-21
Link to the Apple guidelines for explicit


kookster:
Either on the page or on the help docs, here is what I can find to describe what explicit means, which is that it will have a parental advisory on the store:
https://support.apple.com/en-us/HT202005

Issue #155, opened 2016-12-20
Story in a Series should default to the explicit/clean/'' option of parent Series


gcampo88:
If Series is designated Explicit, Story in that Series should default to Explicit too.

sandikbarr:
Adding that the dropdown options should also match. If we have an empty Explicit option for Series, there should be one on Stories too.

PR #182, opened 2016-12-27
default story explicit to series explicit

sandikbarr:
implements #164 and #155

Known issue: #194

### DIFF
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -1,8 +1,13 @@
 <form [ngSwitch]="state">
 
   <publish-fancy-field label="Series Podcast" *ngSwitchCase="'new'">
-    <div class="fancy-hint">You must save your series before you can create
-      a podcast distribution for it.</div>
+    <div class="fancy-hint">The series itself must be created before a podcast distribution
+      can be created for it.</div>
+  </publish-fancy-field>
+
+  <publish-fancy-field label="Series Podcast" *ngSwitchCase="'missing'">
+    <div class="fancy-hint">This series must have audio templates before a podcast distribution
+      can be created for it.</div>
   </publish-fancy-field>
 
   <publish-fancy-field label="Series Podcast" *ngSwitchCase="'creating'">
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -49,7 +49,14 @@
     </publish-fancy-field>
 
     <publish-fancy-field label="Explicit" required=1 [model]="podcast" name="explicit" [select]="explicitOpts">
-      <div class="fancy-hint">Do any of your podcast episodes contain explicit material?</div>
+      <div class="fancy-hint">
+        In accordance with
+        <a [href]="itunesRequirementsDoc"
+           title="Requirements - Podcasts Connect Help">the requirements for iTunes podcast content</a>,
+        do any of your podcast episodes contain
+        <a [href]="itunesExplicitDoc"
+           title="About iTunes Store Parental Advisories - Apple Support">explicit material?</a>
+      </div>
     </publish-fancy-field>
 
     <publish-fancy-field label="Homepage Link" required textinput [model]="podcast" name="link">
--- a/src/app/series/directives/series-podcast.component.ts
+++ b/src/app/series/directives/series-podcast.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnDestroy, DoCheck } from '@angular/core';
 import { Subscription } from 'rxjs';
-import { SeriesModel, DistributionModel, FeederPodcastModel,
+import { SeriesModel, DistributionModel, FeederPodcastModel, AudioVersionTemplateModel,
   TabService, CATEGORIES, SUBCATEGORIES } from '../../shared';
 
 @Component({
@@ -19,12 +19,21 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   tabSub: Subscription;
   state: string;
   series: SeriesModel;
+  templateSub: Subscription;
+  podcastTemplate: AudioVersionTemplateModel;
   distribution: DistributionModel;
   podcast: FeederPodcastModel;
   previewUrl: string;
 
   constructor(tab: TabService) {
-    this.tabSub = tab.model.subscribe(s => this.series = <SeriesModel> s);
+    this.tabSub = tab.model.subscribe(s => {
+      this.series = <SeriesModel> s;
+      this.templateSub = this.series.loadRelated('versionTemplates').subscribe(() => {
+        if (this.series.versionTemplates && this.series.versionTemplates.length > 0) {
+          this.podcastTemplate = this.series.versionTemplates[0];
+        }
+      });
+    });
   }
 
   ngDoCheck() {
@@ -47,6 +56,9 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
       this.state = 'new';
     } else if (this.series) {
       this.state = 'creating';
+      if (!this.podcastTemplate) {
+        this.state = 'missing';
+      }
     } else {
       this.state = null;
     }
@@ -54,6 +66,9 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
 
   ngOnDestroy(): any {
     this.tabSub.unsubscribe();
+    if (this.templateSub) {
+      this.templateSub.unsubscribe();
+    }
   }
 
   loadPodcast() {
@@ -70,9 +85,9 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   }
 
   createDistribution() {
-    let podcast = new DistributionModel(this.series.doc);
-    podcast.set('kind', 'podcast');
-    this.series.distributions.push(podcast);
+    let podcastDist = new DistributionModel({series: this.series.doc, template: this.podcastTemplate.doc});
+    podcastDist.set('kind', 'podcast');
+    this.series.distributions.push(podcastDist);
   }
 
   setSubCategories() {
@@ -91,7 +106,7 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   }
 
   get urlStart() {
-    if (this.podcast) {
+    if (this.podcast && this.podcast.publishedUrl) {
       let parts = this.podcast.publishedUrl.split('/');
       if (parts.length > 2) {
         return parts.slice(0, parts.length - 2).join('/') + '/';
@@ -100,7 +115,7 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   }
 
   get urlEnd() {
-    if (this.podcast) {
+    if (this.podcast && this.podcast.publishedUrl) {
       let parts = this.podcast.publishedUrl.split('/');
       if (parts.length > 2) {
         return '/' + parts.slice(parts.length - 1).join('/');
--- a/src/app/series/directives/series-podcast.component.ts
+++ b/src/app/series/directives/series-podcast.component.ts
@@ -29,6 +29,7 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
       this.series = <SeriesModel> s;
       this.templateSub = this.series.loadRelated('versionTemplates').subscribe(() => {
         if (this.series.versionTemplates && this.series.versionTemplates.length > 0) {
+          // Later we may allow the user to choose which template, but for now they only get one.
           this.podcastTemplate = this.series.versionTemplates[0];
         }
       });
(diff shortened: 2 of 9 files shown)


src/app/story/directives/podcast.component.ts
```diff
@@ -33,12 +37,56 @@ export class PodcastComponent implements OnDestroy {
...
+    } else if (series.unsavedDistribution) {
+      this.defaultExplicitToSeries([series.unsavedDistribution]);
+    }
+  }
+
+  defaultExplicitToSeries(distributions: DistributionModel[]) {
+    let dist = distributions.find(d => d.kind === 'podcast');
+    if (dist) {
+      dist.loadExternal().subscribe(() => {
+        // TODO: how to know which of these versions is the podcast?
```
sandikbarr: In https://github.com/PRX/cms.prx.org/pull/211, looks like the distribution will have an audio version template id. The audio version has a belongs_to relationship with the template. These relationships need to be used to determine which is the podcast audio version in order to apply the default.

src/app/shared/model/audio-version.model.ts
```diff
@@ -11,7 +11,7 @@ export class AudioVersionModel extends BaseModel {
   public id: number;
   public uploads: string = '';
   public label: string;
-  public explicit: string = '';
+  public explicit: string = null;
```
cavis: I'm not sure i follow why this needs to distinguish between `''` and `null`.  Aren't we dealing with 3 states here - blank, Explicit or Clean?

Generally I've been defaulting/decoding values to blank instead of null.  Since the form fields will generally set it back to blank if you delete/unset it in the UI, and I want that to look like it's "unchanged".

src/app/shared/model/audio-version.model.ts
```diff
@@ -39,13 +39,13 @@ export class AudioVersionModel extends BaseModel {
     this.setLabel();
   }
 
-  setLabel() {
+  private setLabel() {
     if (this.doc && this.doc['label']) {
-      this.set('label', this.doc['label']); // probably already set to this
+      this.label = this.doc['label']; // probably already set to this
```
cavis: Are you just trying to avoid triggering the changed-indicator here?

src/app/shared/model/audio-version.model.ts
```diff
@@ -124,9 +127,12 @@ export class AudioVersionModel extends BaseModel {
       case 'Clean':
         data.explicit = 'clean';
         break;
-      default:
+      case '':
```
cavis: Again, I was thinking we should be PUT-ing a `null` to feeder, when it's blank.

src/app/shared/model/audio-version.model.ts
sandikbarr: How can we determine that a value has not been set and to do the override if we default it to one of the available values?

src/app/shared/model/audio-version.model.ts
sandikbarr: Yes, this triggering changed was causing problems. When the story audio version explicit was set to what the series had on a new draft story, it was showing up as changed because of the label (but none of the fields were highlighted orange because it was not explicit that had actually "changed" just had a value assigned as a default.)

src/app/story/directives/podcast.component.ts
```diff
@@ -33,12 +37,56 @@ export class PodcastComponent implements OnDestroy {
...
+        this.defaultDistributionOverrides(new SeriesModel(account, this.story.parent, false));
+      } else if (!this.story.isNew && this.story.doc.has('prx:series')) {
+        this.story.doc.follow('prx:series').subscribe(series => this.defaultDistributionOverrides(new SeriesModel(account, series, false)));
+      }
+    });
+  }
+
+  defaultDistributionOverrides(series: SeriesModel) {
+    if (series.doc && series.doc.count('prx:distributions')) {
+      series.doc.followItems('prx:distributions').subscribe(distributions => {
```
cavis: This _should_ be easier someday (after #156) ... which also allows you to subscribe to the loading-of-related, and do stuff after they return.

src/app/shared/model/audio-version.model.ts
sandikbarr: So when story audio version explicit is null (even if set to null by the user because they don't want a parental advisory tag on this story) but the series has explicit or clean, it will always override the story. Unless I'm missing some other indicator here that explicit was not set.

src/app/story/directives/podcast.component.ts
```diff
@@ -33,12 +37,56 @@ export class PodcastComponent implements OnDestroy {
...
+  defaultExplicitToSeries(distributions: DistributionModel[]) {
+    let dist = distributions.find(d => d.kind === 'podcast');
+    if (dist) {
+      dist.loadExternal().subscribe(() => {
+        // TODO: how to know which of these versions is the podcast?
+        this.story.versions.forEach(v => {
+          if (dist.podcast && v.explicit === null) {
+            // TODO: how to know that audio version explicit has not been set?
+            // audio version explicit is empty string when it has not been set, which is a valid value
+            v.explicit = dist.podcast.explicit;
```
sandikbarr: There's gonna be a useAsOriginal flag on BaseModel.set() soon, use that to override the initial empty string value instead of plain assignment. Having the override value as the original value allows us to use changed() to determine if the value should be overridden on draft stories.

src/app/shared/model/distribution.model.spec.ts
```diff
@@ -8,37 +8,46 @@ describe('DistributionModel', () => {
   let series = cms.mock('prx:series', {id: 'series1'});
   let fooDist = series.mock('prx:distributions', {id: 'dist1', kind: 'foo'});
   let podDist = series.mock('prx:distributions', {id: 'dist2', kind: 'podcast', url: podcastUrl});
+  let seriesVersionTemplates = series.mockItems('prx:audio-version-templates', []);
+  let seriesVersionFileTemplates = series.mockItems('prx:audio-file-templates', [{}]);
+  let seriesVersionFileTemplateItems = seriesVersionFileTemplates[0].mockItems('prx:items', [{}]);
+  let fooDistTemplate = fooDist.mock('prx:audio-version-template', {id: 'templateFoo'});
+  let fooDistFileTemplate = fooDistTemplate.mock('prx:audio-file-templates', {id: 'fileFoo'});
+  let fooDistFileTemplateItems = fooDistFileTemplate.mockItems('prx:items', [{}]);
```
sandikbarr: Something about this is mocked incorrectly, throwing error `Expected mocked array at prx:items - got object`

src/app/story/directives/podcast.component.ts
```diff
@@ -33,12 +37,53 @@ export class PodcastComponent implements OnDestroy {
...
+    });
+  }
+
+  defaultExplicitToSeries(distributions: DistributionModel[]) {
+    let dist = distributions.find(d => d.kind === 'podcast');
+    dist.loadRelated('versionTemplate').subscribe(() => {
+      dist.loadExternal().subscribe(() => {
+        this.story.loadRelated('versions').subscribe(() => {
+          this.story.versions.forEach(v => {
+            // well crap, new story's version doc is null
```
sandikbarr: @cavis This is pretty much where I keep getting stuck. I'm this far along, but since the story is new, its versions do not have a doc. (I do intend to add the relation on audio versions to audio version templates and loadRelated, just working with what I've got for now.)

src/app/shared/model/distribution.model.spec.ts
cavis: Ah I see.  So that `mockItems` method actually takes care of doing the prx items for you.  So long syntax (as you're doing above) would be:

```
let fooDistTemplate = fooDist.mock('prx:audio-version-template', {id: 'templateFoo'});
let fooDistFileTemplate = fooDistTemplate.mock('prx:audio-file-templates', {});
let fooDistFileTemplateItems = fooDistFileTemplate.mockList('prx:items', [{}]);
```

The shorthand combines those 2 last lines:

```
let fooDistTemplate = fooDist.mock('prx:audio-version-template', {id: 'templateFoo'});
let fooDistFileTemplates = fooDistTemplate.mockItems('prx:audio-file-templates', [{id: 'fileFoo'}]);
```

src/app/story/directives/podcast.component.ts
cavis: Okay, so this is actually really close.  (I ended up going down some rabbit holes debugging, because I had some pre-this-PR distributions that didn't have any audio_version_template set).

So the `VersionModel` may not have a doc, but it does have a `template` property, which will be non-null if it was init'd with an `AudioVersionTemplateModel`.  So you can just check `if (v.template && v.template.id === dist.versionTemplate.id)` or something.

AudioVersionModel is a bit strange in that its AudioVersionTemplateModel isn't actually related to it.  It's assigned to it from the constructor, because new AudioVersionModels need to know which template they are for right away.

sandikbarr: Thanks! Most of our other ids are in the doc, so I wasn't looking for one in the model.

I was thinking I should add the relation for AudioVersionModel to AudioVersionTemplate though for the loadRelated support. Is that a bad idea?

cavis: So currently, the only place you see `new AudioVersionModel()` is in the `StoryModel`.  And it always inits the new/existing versions with their template (the `vdoc.follow('prx:audio-version-template')` is in there).  So you don't need to add it as a relation.

That said, it would be nice to revisit/clean-up how many of these related are loaded (and maybe make them all load on-demand rather than preloading).  But I'd prefer to knock that out as an independent PR, since it touches so many things.

src/app/series/directives/series-podcast.component.html
```diff
@@ -1,8 +1,13 @@
 <form [ngSwitch]="state">
 
   <publish-fancy-field label="Series Podcast" *ngSwitchCase="'new'">
-    <div class="fancy-hint">You must save your series before you can create
-      a podcast distribution for it.</div>
+    <div class="fancy-hint">The series itself must be created before a podcast distribution
+      can be created for it.</div>
```
sandikbarr: I changed this language because a) I thought it should say create to match the button rather than save and b) it's said that using "You" or "You must" in alerts or error messages can be interpreted as insulting or condescending.

cavis: Okay fine.  But I'm still going to insult our users every night in my diary.

src/app/series/directives/series-podcast.component.html
```diff
@@ -1,8 +1,13 @@
...
+  </publish-fancy-field>
+
+  <publish-fancy-field label="Series Podcast" *ngSwitchCase="'missing'">
+    <div class="fancy-hint">This series must have audio templates before a podcast distribution
```
cavis: I just started doing similar logic in #185, and realized i should just merge this instead.

src/app/series/directives/series-podcast.component.ts
```diff
@@ -44,13 +56,19 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
...
+      }
     } else {
       this.state = null;
     }
   }
 
   ngOnDestroy(): any {
     this.tabSub.unsubscribe();
+    if (this.templateSub) {
+      this.templateSub.unsubscribe();
```
cavis: You may not actually need this.  HalDocs (underlying the `loadRelated()` call) emit a single object/array and then `finish()`, so you generally don't need to track/unsubscribe from them.  The `tabSub` is a bit different, as it's going to outlive this single tab.

That said ... this doesn't actually hurt anything.

src/app/series/directives/series-podcast.component.ts
```diff
@@ -66,9 +84,9 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   }
 
   createDistribution() {
-    let podcast = new DistributionModel(this.series.doc);
-    podcast.set('kind', 'podcast');
-    this.series.distributions.push(podcast);
+    let podcastDist = new DistributionModel({series: this.series.doc, template: this.podcastTemplate.doc});
```
cavis: I guess there's kind of a race condition here between loading the `versionTemplates` and the user clicking "Create Podcast".  But I'm not worried about it.

src/app/shared/model/audio-version.model.ts
```diff
@@ -44,13 +44,13 @@ export class AudioVersionModel extends BaseModel implements HasUpload {
     this.setLabel();
   }
 
-  setLabel() {
+  private setLabel() {
     if (this.doc && this.doc['label']) {
-      this.set('label', this.doc['label']); // probably already set to this
+      this.set('label', this.doc['label'], true); // probably already set to this
```
cavis: This all feels right to me.  New AudioVersions should look like they haven't `changed()`, so we should force-set the labels on them.

src/app/shared/model/distribution.model.spec.ts
```diff
@@ -8,37 +8,44 @@ describe('DistributionModel', () => {
...
-    let dist = new DistributionModel(series, fooDist);
+    let dist = new DistributionModel({series, distribution: fooDist});
     expect(dist.kind).toEqual('foo');
     expect(dist.RELATIONS).toContain('podcast');
     expect(dist.podcast).toBeNull();
   });
 
   it('creates a new podcast for new distributions', () => {
-    let dist = new DistributionModel(series);
+    let dist = new DistributionModel({series});
```
cavis: I didn't know you could do this in typescript!

Issue #124, opened 2016-12-07
Feed story ordering


cavis:
Somewhere in Publish, show the order of the stories as they will appear in the feed. Maybe a tab on series-edit?

sandikbarr:
Seems like we also want a sort by feed order on the search page

cavis:
Sorting by "published_at" will give you the order of the feed-items.  Though it's a bit confusing, as I don't think we display the published date anywhere in the "card" UI.

I was thinking of this as a tab on the Series page, which just fetched all the Stories in that Series, and gave you a real simple list of them ordered by published-at.  (The order of the items in your feed).  Make sense?

Or maybe this is entirely useless.

sandikbarr:
That makes sense, yeah 👍

sandikbarr:
Stories will have both a public/private flag as well as a publishedAt date. The tab should represent all stories within a series. Stories without a publishedAt date will be at the top of the sorted list. Stories that do have a publishedAt date but are not in the public feed (ones that are private or with publishedAt dates in the future) should have some indicator in the UI to represent their status, 
maybe a different background color or something.

PR #183, opened 2016-12-28
Show podcast feed story order

gcampo88:
Added a tab on Series edit that shows the stories for a podcast in the order they'll show up in the feed (#124), and will sort based on public / to-be-public / private status.

### DIFF
--- a/src/app/series/directives/series-feed.component.ts
+++ b/src/app/series/directives/series-feed.component.ts
@@ -82,7 +82,7 @@ export class SeriesFeedComponent implements OnDestroy {
           .subscribe((docs) => {
             this.isLoaded = true;
             docs.forEach((doc) => {
-                let story = new StoryModel(this.list.parent, doc, false);
+                let story = new StoryModel(this.series.doc, doc, false);
                 if (!story.publishedAt) {
                   this.privateStories.push(story);
                   return;
--- a/src/app/series/directives/series-feed.component.ts
+++ b/src/app/series/directives/series-feed.component.ts
@@ -1,12 +1,12 @@
 import { Component, OnDestroy } from '@angular/core';
 import { Subscription } from 'rxjs';
-import { DistributionModel, SeriesModel, StoryModel, TabService } from '../../shared';
+import { SeriesModel, StoryModel, TabService } from '../../shared';
 
 @Component({
   template: `
   <div>
     <publish-spinner *ngIf="!isLoaded"></publish-spinner>
-    <section *ngIf="series && list">
+    <section *ngIf="series">
       <div class="hint" *ngIf="noStories">
         You have no published stories in this series.
       </div>
@@ -56,17 +56,13 @@ export class SeriesFeedComponent implements OnDestroy {
   publicStories: StoryModel[] = [];
   futurePublicStories: StoryModel[] = [];
   privateStories: StoryModel[] = [];
-  list: DistributionModel;
   tabSub: Subscription;
 
   constructor(tab: TabService) {
     this.noStories = false;
     this.tabSub = tab.model.subscribe((s: SeriesModel) => {
       this.series = s;
-      s.loadRelated('distributions').subscribe((dists) => {
-        this.list = dists[0];
-        this.sortStories();
-      });
+      this.sortStories();
     });
   }
 
--- a/src/app/series/directives/series-feed.component.ts
+++ b/src/app/series/directives/series-feed.component.ts
@@ -6,9 +6,9 @@ import { DistributionModel, SeriesModel, StoryModel, TabService } from '../../sh
   template: `
   <div>
     <publish-spinner *ngIf="!isLoaded"></publish-spinner>
-    <section *ngIf="series && podcast">
+    <section *ngIf="series && list">
       <div class="hint" *ngIf="noStories">
-        You have no published stories in this podcast.
+        You have no published stories in this series.
       </div>
 
       <section *ngIf="!noStories">
@@ -29,7 +29,7 @@ import { DistributionModel, SeriesModel, StoryModel, TabService } from '../../sh
         </ul>
 
         <ul *ngIf="publicStories.length">
-          <h4>Public feed</h4>
+          <h4>Public</h4>
           <li *ngFor="let s of publicStories">
             <h5><a [routerLink]="['/story', s.id]">{{s.title}}</a></h5>
             <p class="public">Published {{s.pubDate}}.</p>
@@ -50,20 +50,22 @@ export class SeriesFeedComponent implements OnDestroy {
   publicStories: StoryModel[] = [];
   futurePublicStories: StoryModel[] = [];
   privateStories: StoryModel[] = [];
-  podcast: DistributionModel;
+  list: DistributionModel;
   tabSub: Subscription;
 
   constructor(tab: TabService) {
     this.noStories = false;
     this.tabSub = tab.model.subscribe((s: SeriesModel) => {
       this.series = s;
-      this.podcast = s.distributions.filter(dist => dist.kind === 'podcast')[0];
-      this.sortStories();
+      s.loadRelated('distributions').subscribe((dists) => {
+        this.list = dists[0];
+        this.sortStories();
+      });
     });
   }
 
   sortStories() {
-    this.podcast
+    this.list
         .parent
         .followItems('prx:stories', { sorts: 'updated_at:desc' })
         .subscribe((docs) => {
@@ -73,7 +75,7 @@ export class SeriesFeedComponent implements OnDestroy {
             return;
           }
           docs.forEach((doc) => {
-              let story = new StoryModel(this.podcast.parent, doc, false);
+              let story = new StoryModel(this.list.parent, doc, false);
               if (!story.publishedAt) {
                 this.privateStories.push(story);
                 return;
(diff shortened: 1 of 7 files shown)


src/app/series/directives/series-feed.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  futurePublicStories: StoryModel[] = [];
+  privateStories: StoryModel[] = [];
+  podcast: DistributionModel;
+  tabSub: Subscription;
+
+  constructor(tab: TabService) {
+    this.noStories = false;
+    this.tabSub = tab.model.subscribe((s: SeriesModel) => {
+      this.series = s;
+      this.podcast = s.distributions.filter(dist => dist.kind === 'podcast')[0];
```
cavis: This works, because distributions are `_embedded` in the series from CMS.  But it might be safer to wait for them to load:

```
s.loadRelated('distributions').subscribe(dists => this.podcast = dists.find(d => d.kind === 'podcast'));
```

src/app/series/directives/series-feed.component.ts
```diff
@@ -0,0 +1,89 @@
...
+      <section *ngIf="!noStories">
+        <ul *ngIf="privateStories.length">
+          <h4>Private</h4>
+          <li *ngFor="let s of privateStories">
+            <h5><a [routerLink]="['/story', s.id]">{{s.title}}</a></h5>
+            <p class="private">No publication date set.</p>
+          </li>
+        </ul>
+
+        <ul *ngIf="futurePublicStories.length">
```
cavis: Since this could potentially take awhile to load, might be nice to show a loading-spinner here.

src/app/series/directives/series-feed.component.ts
```diff
@@ -0,0 +1,89 @@
...
+      this.series = s;
+      this.podcast = s.distributions.filter(dist => dist.kind === 'podcast')[0];
+      this.sortStories();
+    });
+  }
+
+  sortStories() {
+    this.podcast
+        .parent
+        .followItems('prx:stories', { sorts: 'updated_at:desc' })
```
cavis: Can CMS not just sort by `published_at:desc` to start with?  Or would that not include `published_at = null` stories?

src/app/series/directives/series-feed.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.noStories = false;
+    this.tabSub = tab.model.subscribe((s: SeriesModel) => {
+      this.series = s;
+      this.podcast = s.distributions.filter(dist => dist.kind === 'podcast')[0];
+      this.sortStories();
+    });
+  }
+
+  sortStories() {
+    this.podcast
```
cavis: Rather than calling `this.podcast.parent.followStories` (it's possible this series as no podcast-distribution, in which case you'll get an error) ... you could just look at the `series.doc.followItems(...)`.

src/app/series/directives/series-feed.component.ts
cavis: And the more I think about it ... I'm not sure this tab should care if you have a podcast-distribution or not.  Even on a non-podcast Series, it would be useful to see this list.

src/app/series/series.component.html
```diff
@@ -31,6 +31,7 @@ <h1 class="hero-title" *ngIf="!id">Create Series</h1>
     <a routerLinkActive="active" [routerLinkActiveOptions]="{exact:true}" [routerLink]="base">Basic Info</a>
     <a routerLinkActive="active" [routerLink]="[base, 'templates']">Audio Templates</a>
     <a routerLinkActive="active" [routerLink]="[base, 'podcast']">Podcast Distribution</a>
+    <a routerLinkActive="active" [routerLink]="[base, 'feed']">Stories in Feed</a>
```
cavis: Should probably hide this tab when `series.isNew`.  I'm also leaning towards not using the word "Feed" here, in case this series isn't actually a podcast.  Thoughts?

src/app/series/series.component.html
```diff
@@ -31,6 +31,7 @@ <h1 class="hero-title" *ngIf="!id">Create Series</h1>
...
   </nav>
   <div class="extras">
     <a *ngIf="storyCount > 0" [routerLink]="['/search', { tab: 'stories', seriesId: id }]">{{storyCount}} {{storyNoun}} &raquo;</a>
```
cavis: Wondering if you should remove this link, and maybe put it on your new tab somewhere.  Your tab-label could also use this count, if you want it to be a bit more descriptive.

src/app/series/directives/series-feed.component.ts
cavis: I also think there's a problem with paging here - CMS will by default only return the first 10, with links to the next/first/last pages.  So here you may need to do something like (don't quote me here - guessing on some syntax):

```
let total = this.series.doc.total('prx:stories');
if (total > 0) {
  this.series.followItems('prx:stories', {per: total, sorts: 'published_at:desc'}).subscribe(stories => {
    // do stuff here
  });
}
```

You'll have to upgrade the `HalDoc.total()` method to optionally accept a `rel` param, the same way the `count()` method does.  And this could be a potentially long-running query, depending on how many stories there are.  (I tried it in prod on a total=318 series, and it was real slow).  But we can optimize that later!  (Perhaps getting CMS to not embed so many things in the respones).

src/app/series/directives/series-feed.component.ts
gcampo88: :+1:

src/app/series/directives/series-feed.component.ts
gcampo88: :+1:

src/app/series/series.component.html
gcampo88: sounds good!

src/app/series/directives/series-feed.component.ts
gcampo88: 👍

src/app/series/directives/series-feed.component.ts
gcampo88: Why not just use `Haldoc.count(rel)` for this? `let total = this.series.doc.count('prx:stories');` ?

src/app/series/series.component.html
gcampo88: :+1:

src/app/series/directives/series-feed.component.ts
cavis: The count is actually the number of docs "in this page".  For instance: https://cms.prx.org/api/v1/series/33698/stories?per=12

gcampo88: aha. thanks! will rework total..

gcampo88: so while walking Luna I realized I could just copy the storyCount-obtaining code from `series.component.ts` in order to get the total # of stories for this new tab -- but looks like it is using count() instead of total() too. 
```
if (series) {
      this.storyCount = series.count('prx:stories');
      this.storyNoun = this.storyCount === 1 ? 'Story' : 'Stories';
} 
```
Does this work for your series with 318 stories in it?

cavis: Hmm... yes, that one is going to be the total.  But there's some funny stuff going on here.

If you look at this [series doc](https://cms.prx.org/api/v1/series/33698) ... the `_embedded.prx:stories` has count = 10.  But the `_links.prx:stories` has count = 318.  So if you look at the embedded, it's going to be the page count.  But the links will have the total.

So that makes me thing that `series.count('prx:stories')` is what you actually want.  I'd want to test that to confirm, though.  It's a bit confusing.

gcampo88: Agreed. One other odd thing is -- I think the `total` and `count` number of stories in the _embedded may only be those that are currently published, while _links is showing all stories regardless of pub status... ? Could be coincidence but I have a series of 12 stories, and here's what I'm seeing: 
```javascript
this.series.doc._embedded['prx:stories'].count //= 4
this.series.doc._embedded['prx:stories'].total //= 4
this.series.doc._links['prx:stories'].count //= 12
```

Could it be because the _embedded uses the non-authorized uri
`href: "/api/v1/series/36512/stories"`
and _links uses the authorized one `href: "/api/v1/authorization/series/36512/stories{?page,per,zoom,filters,sorts}"`
?

cavis: Interesting.  Sounds like a bug in CMS, if it's embedding something other than what following the link would have returned.

gcampo88: i think that may be the case-- here's a screenshot from debugger in the Chrome console comparing `_embedded['prx:stories']` to `_links['prx:stories']` for series 36512: 
screen shot 2017-01-05 at 4 15 37 pm

gcampo88:
@cavis think I've addressed everything!

src/app/series/directives/series-feed.component.ts
```diff
@@ -0,0 +1,104 @@
...
+  privateStories: StoryModel[] = [];
+  list: DistributionModel;
+  tabSub: Subscription;
+
+  constructor(tab: TabService) {
+    this.noStories = false;
+    this.tabSub = tab.model.subscribe((s: SeriesModel) => {
+      this.series = s;
+      s.loadRelated('distributions').subscribe((dists) => {
+        this.list = dists[0];
```
cavis: Seems like this `list` (DistributionModel) isn't actually used for anything.  And as much as I like to see `loadRelated` being used ... you can probably delete this whole block and just do `sortStories()` after setting `this.series = s`.

gcampo88:
good call! Should be fixed now.

Issue #166, opened 2016-12-21
Story, change published_at


kookster:
- [ ] Instead of just a publish button and boolean publish status, set published_at date
- [ ] Also be able to set the date in the past or future. Stories with published_at set in the future will be visible and will be added to the podcast feed when the date becomes current.

sandikbarr:
Should use a date or date and time picker. Times will not be precise, 30 minute increments would be sufficient. Possibly implement by changing Publish button to a Publish On datepicker control.

sandikbarr:
Andrew Kuklewicz [9:02 AM] 
could be a publish on, with a prominent "now", as that is usually what people want
and currently positioned above the main form div  
which probably still makese sense

Issue #154, opened 2016-12-19
Publish button handling


cavis:
Since we call things "public" and "private" on the dashboard, we should rethink the current publish/unpublish buttons on the story page.

Questions to answer:

- Keep setting it "above the fold"?  (Meaning it's a button/action rather than a localStorage-able state on the Story resource).  Or just make it an element on the form?
- How does this play with explicitly setting the `published_at` date to influence your podcast item ordering?

I was leaning more towards keeping this an action/button, with a bootstrap-style [button dropdown](http://getbootstrap.com/components/#btn-dropdowns).  This could allow you to alter it as a boolean, or explicitly set the date.  And calling it out above-the-fold prevents users from having unsaved-changes to the public/private state.

sandikbarr:
Related to #166

cavis:
I'm calling this a duplicate of #166, since this will be solved there.

PR #184, opened 2016-12-29
set releasedAt date

sandikbarr:
This is a bit of a change from the original #166 for publishedAt handling. This change should allow the user to set the releasedAt date as a regular form field that is not required. If the releasedAt date is set, CMS will use it for the publishedAt date; otherwise, current date time.


src/app/shared/date/datepicker.component.ts
```diff
@@ -0,0 +1,48 @@
...
+      field: this.input.nativeElement,
+      format: DatepickerComponent.FORMAT,
+      theme: 'triangle-theme',
+      onSelect: () => {
+        this.onDateChange.emit(this.picker.getDate());
+      }
+    };
+    if (this.date) {
+      // publishedAt is declared as a Date. why is it a string?
+      let defaultDate = new Date(this.date.valueOf());
```
sandikbarr: TODO: just use moment since I added it for formatting support

src/assets/images/calendar.svg
```diff
@@ -0,0 +1 @@
+<svg width="1792" height="1792" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg"><path d="M192 1664h288v-288h-288v288zm352 0h320v-288h-320v288zm-352-352h288v-320h-288v320zm352 0h320v-320h-320v320zm-352-384h288v-288h-288v288zm736 736h320v-288h-320v288zm-384-736h320v-288h-320v288zm768 736h288v-288h-288v288zm-384-352h320v-320h-320v320zm-352-864v-288q0-13-9.5-22.5t-22.5-9.5h-64q-13 0-22.5 9.5t-9.5 22.5v288q0 13 9.5 22.5t22.5 9.5h64q13 0 22.5-9.5t9.5-22.5zm736 864h288v-320h-288v320zm-384-384h320v-288h-320v288zm384 0h288v-288h-288v288zm32-480v-288q0-13-9.5-22.5t-22.5-9.5h-64q-13 0-22.5 9.5t-9.5 22.5v288q0 13 9.5 22.5t22.5 9.5h64q13 0 22.5-9.5t9.5-22.5zm384-64v1280q0 52-38 90t-90 38h-1408q-52 0-90-38t-38-90v-1280q0-52 38-90t90-38h128v-96q0-66 47-113t113-47h64q66 0 113 47t47 113v96h384v-96q0-66 47-113t113-47h64q66 0 113 47t47 113v96h128q52 0 90 38t38 90z"/></svg>
```
sandikbarr: This is a font icon. Are we able to add more to what we have with fontello or would we rather not favor the use of font icons?

src/app/story/directives/hero.component.ts
```diff
@@ -93,7 +111,51 @@ export class StoryHeroComponent implements OnInit, OnChanges {
...
+    return this.pendingPublishedAt.getMinutes() <= 30 ? '30' : '00';
+  }
+
+  set pendingPublishedAtMinutes(minutes: string) {
+    this._pendingPublishedAt.setMinutes(Number(minutes));
+  }
+
   togglePublish() {
+    if (!this.story.publishedAt) {
+      this.story.set('publishedAt', this._pendingPublishedAt);
```
sandikbarr: This is a bad deal because it means the story has changed(). Not sure what the pattern should be to set and persist this value. But I think we first need to resolve what the requirements are for #154.

(Also not sure if the user can set publishedAt when the story has already been published or if dates in the past are allowed.)

src/app/shared/model/story.model.ts
```diff
@@ -103,6 +103,7 @@ export class StoryModel extends BaseModel {
     data.shortDescription = this.shortDescription;
     data.descriptionMd = this.description;
     data.tags = this.splitTags();
+    data.publishedAt = this.publishedAt;
```
sandikbarr: TODO: remove from SETABLE if not using model.set()

src/assets/images/calendar.svg
cavis: Yes - you can import the current setup into fontello, add some more to it, and re-export it.

I think you'll also have to update a couple asset-paths in the css, after export.

src/app/shared/hero/hero.component.css
```diff
@@ -36,7 +36,7 @@
 }
 .toolbar section {
   display: flex;
-  align-items: center;
```
sandikbarr: I adjusted the toolbar layout to align to the bottom. The last saved and published at info not being aligned bothered me. I also opened bug #199 for 2 line titles.

src/app/shared/date/timepicker.component.ts
```diff
@@ -0,0 +1,53 @@
+import { Component, Input, Output, EventEmitter } from '@angular/core';
```
sandikbarr: This time picker could probably also be implemented as an add-on to the datepicker. We could also use the pikaday fork that adds time picker support, not sure if it is well maintained though.

src/app/story/directives/hero.component.ts
```diff
@@ -46,6 +47,12 @@ import { StoryModel } from '../../shared';
              [working]="story.isPublishing" (click)="togglePublish()" orange=1>
             {{story.publishedAt ? 'Unpublish' : 'Publish'}}
           </publish-button>
+          <p *ngIf="!story.changed() && !story.publishedAt">
+            Will be queued up for publishing {{story.releasedAt ? ' at ' + formatDate(story.releasedAt) : 'immediately'}}
```
sandikbarr: I don't love this language. It's kind of long and maybe not the best. Any better suggestions? 
 - Does it need to say anything at all when the release date is not set? Maybe it should just say so when the release date is set since that is the anomaly. Otherwise, seems obvious that Publish means now without saying so.
 - To shorten it a bit, maybe show date without time? Time may not be necessary here since it is on the page with the releasedAt date

src/assets/fontello/LICENSE.txt
```diff
@@ -3,7 +3,7 @@ Font license info
```
sandikbarr: Fontello stuff doesn't really need PRed. I regenerated it to add the calendar icon.

src/app/story/directives/basic.component.ts
```diff
@@ -48,6 +48,17 @@ import { StoryModel, TabService } from '../../shared';
       tiptext="More is better!" >
         <div class="fancy-hint">A comma-separated list of tags relevant to the content of your piece.</div>
       </publish-fancy-field>
+      
+      <publish-fancy-field label="Release Date">
+        <div class="fancy-hint">If the release date is not set manually,
+          the current date will be used for publishing.</div>
+        <publish-datepicker
```
cavis: Am I missing some CSS?  When I focus this field, the popup seems to have a transparent background -> http://cav.is/s/5vio7.jpg

src/app/story/directives/basic.component.ts
```diff
@@ -48,6 +48,17 @@ import { StoryModel, TabService } from '../../shared';
       tiptext="More is better!" >
         <div class="fancy-hint">A comma-separated list of tags relevant to the content of your piece.</div>
       </publish-fancy-field>
+      
+      <publish-fancy-field label="Release Date">
```
cavis: Currently not seeing any changed/invalid indicators for this field(s).  Would it be worth calling out `date` as an additional `@Input`ed subtype in FancyFieldComponent?  Or alternately: explicitly bind changed/invalid classes to the input elements.

src/assets/fontello/css/fontello.css
```diff
@@ -1,11 +1,11 @@
 @font-face {
   font-family: 'fontello';
-  src: url('/assets/fontello/font/fontello.eot?1138238');
-  src: url('/assets/fontello/font/fontello.eot?1138238#iefix') format('embedded-opentype'),
-       url('/assets/fontello/font/fontello.woff2?1138238') format('woff2'),
-       url('/assets/fontello/font/fontello.woff?1138238') format('woff'),
-       url('/assets/fontello/font/fontello.ttf?1138238') format('truetype'),
-       url('/assets/fontello/font/fontello.svg?1138238#fontello') format('svg');
+  src: url('../font/fontello.eot?92701275');
```
cavis: For some reason i thought these would need absolute urls... but apparently not!

src/app/shared/hero/hero.component.css
cavis: This does make the buttons look a bit funny on the Series page though (the toolbar is a shared component) -> http://cav.is/s/gystn.jpg

cavis: That extra line of text on the Story page also messes up the toolbar-affixing a bit.  (I think that affix stuff is currently based on a fixed-height toolbar).

src/app/story/directives/hero.component.ts
cavis: I know it was in the requirements, but i'm not a huge fan of adding more stuff to the toolbar.  Sort of clutters the place up, and messes with whitespace (http://cav.is/s/b1gbi.jpg).

Not going to hold things up based on my opinion alone... but as an alternative to displaying this hint here:

- If publishedAt is null and releasedAt is null-or-in-the-past, button text says `Publish Now`
- If publishedAt is null and releasedAt is in-the-future, button says `Publish In X Hours` (or days, or minutes, etc).  Just a short preview of the distance between `now` and `releasedAt`
- If publishedAt is in-the-future, button has the "Private" color from the dashboard and says `Publishing in X Hours`
  - On hover, button says `Unpublish` and lets you set `publishedAt` back to null again
- If publishedAt is in-the-past, button has the "Public" color from the dashboard and says `Published`
  - On hover, button says `Unpublish` and lets you set `publishedAt` back to null again

cavis: Or another alternative: use the current `Last saved at` status text to indicate whatever hints we need about published status.

cavis: Or use a popup tip, the same sort you get when hovering over the disabled `Save` button when your changes are invalid.

cavis: These may be bad ideas.  Just brainstorming alternatives.

src/app/story/directives/basic.component.ts
sandikbarr: That CSS is loaded by the angular-cli, so if you don't restart the app, it doesn't get loaded.

src/app/story/directives/hero.component.ts
gcampo88: I like @cavis's ideas if we want to change the button text -- otherwise, if we want to stick with hint text below the button, I'd vote for brevity over explanation of process-- aka, just say "Will be published on {{story.releasedAt}}" (skip the mention of queuing, and also skip the time). One thing with the text as is is that if `releasedAt` is deliberately set in the past, the text sounds a little funny since it is future-focused. Not sure if this way would be any better.

src/app/story/directives/basic.component.ts
sandikbarr: I added changed support, but not invalid. Typing garbage into the input box does not change the model, but it should show invalid.

src/app/story/directives/hero.component.ts
```diff
@@ -20,7 +21,8 @@ import { StoryModel } from '../../shared';
       <div class="hero-info" *ngIf="story">
         <h2>{{story.title || '(Untitled)'}}</h2>
         <p *ngIf="story?.isNew">Not saved</p>
-        <p *ngIf="!story?.isNew">Last saved at {{story.updatedAt | timeago}}</p>
+        <p *ngIf="!story?.isNew && !story?.publishedAt">Last saved at {{story.updatedAt | timeago}}</p>
+        <p *ngIf="story?.publishedAt">{{publishedOnText}}</p>
```
sandikbarr: Hoping we can settle on this for now and make any adjustments needed after user feedback.

gcampo88:
I'm having an issue I wonder if you're experiencing too. Steps to reproduce:
1) Set a released date (I did mine in the future, not sure it will matter as long as it's not null) for a story. Click publish. Text updates to "Will be published on {correct future date}". 
2) Unpublish. Text reverts to "Last saved". Change the released at date. Click save. 
3) Click publish. Text updates to "Will be published on... " but uses the original released_at date as the publication date, instead of the new one. And meanwhile the Release Date field has also reverted to the previous value. 

Similar prob is if I change the Release Date while the story is already published, click Save, and then click Unpublish, the Release Date reverts to its original value. 

Are you having this issue?

Issue #168, opened 2016-12-21
Story, Podcast Distribution fields


kookster:
- [ ] Show and make overridable the GUID - overwrite by writing to `feeder.episode.original_guid`
- [ ] Show and make overridable the author, name and email, starting with podcast author, which defaults to prx user
- [ ] Show the enclosure url, not editable

sandikbarr:
Sandi Barr [10:07 AM]  
My assumption on these two https://github.com/PRX/publish.prx.org/issues/168 and https://github.com/PRX/publish.prx.org/issues/167 is that they've got feeder dependencies. LMK if I'm incorrect and those are ready to be worked.
Ryan Cavis [10:26 AM]  
there’s probably one TODO there to make CMS send the “defaults” to Feeder
but i think you should be able to get/set that data from feeder
the current FeederPodcastModel is just getting/putting to the podcasts endpoint
https://feeder.prx.org/api/v1/podcasts/23
you can see in roman’s setup, many more of the fields are there than in a brand-new-podcast you create
but you should be able to send those fields with your new Podcast
Andrew Kuklewicz [10:29 AM]  
you can do both #167 and #168, they use fields that are in feeder, but are already available in the feeder api
the main thing that is missing is for the author (not the guid or enclosure url), which is to make sure once overridden in feeder
it doesn't get rewritten by an update from cms, which currently sends all its updates via asynch messages
to update related feeder entries
but, you can do all the publish work for 167 and 168 w/o that override tracking being in place
Ryan Cavis [10:31 AM]  
ah, right
Andrew Kuklewicz [10:32 AM]  
we also should send more of the defaults to feeder on story create
but that also is behind the scenes, won't affect how these fields are made editable on publish
just how soon a default value shows up in them, either immediately, or after the asynch pudate

sandikbarr:
Related: https://github.com/PRX/feeder.prx.org/issues/161

sandikbarr:
How should I get the episode by prx story from feeder? Do we need to add a story or prxUri filter_param or is there support for this already that I'm just not seeing?

PR #185, opened 2016-12-30
Story Podcast Distribution

sandikbarr:
#168


src/app/core/cms/abstract.service.ts
```diff
@@ -0,0 +1,68 @@
+import { Observable, ReplaySubject } from 'rxjs';
+import { Http } from '@angular/http';
+import { HalDoc } from './haldoc';
+import { HalObservable } from './halobservable';
+import { HalRemote } from './halremote';
+
+export abstract class AbstractService {
```
sandikbarr: CmsService could/should also extend AbstractService (or BaseService or whatever we want to call it)

src/app/shared/model/feeder-episode.model.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { cms as cms, cms as feeder } from '../../../testing';
```
sandikbarr: MockCmsService could just be plain MockService

sandikbarr: MockCmsService could just be plain MockService

src/app/shared/model/feeder-episode.model.ts
```diff
@@ -0,0 +1,68 @@
...
+    return {};
+  }
+
+  decode() {
+    // hmm, episode id is string but Haldoc id is number and making it number | string throws all kinds of errors
+    this.id = '' + this.doc['id'];
+    this.guid = this.doc['guid'];
+    // TODO: needs to default to account if not set,
+    //  should this be done here (with follow on story prx:account like with FeederPodcastModel)
+    //  or where we're using this model for the story?
```
sandikbarr: Is this defaulting to account name and email a model responsibility? Should get via this.story.follow('prx:account') like [FeederPodcastModel](https://github.com/PRX/publish.prx.org/blob/master/src/app/shared/model/feeder-podcast.model.ts#L61)?

src/app/core/core.env.ts
```diff
@@ -7,6 +7,7 @@ const DEFAULTS = {
   CMS_TTL: 1, // 1 second
   CMS_ROOT_TTL: 3600, // 1 hour
   CMS_USE_LOCALSTORAGE: false,
+  FEEDER_HOST: 'https://feeder.prx.org',
```
cavis: Does Publish need to actually know where Feeder is?  CMS is already configured to know, and returns absolute links to Feeder (or any other PRX Distribution Service) in the `distribution` / `story_distribution` representers.  The idea being that Publish can re-use the same ID auth token to authenticate against any PRX service, and they're all HAL, so we can navigate across them seamlessly.

cavis:
This is currently blocked by PRX/cms.prx.org#214 ... need that to be able to navigate links from the `story` -> `story_distribution where kind=podcast` -> `feeder episode`.

src/app/core/core.env.ts
sandikbarr: Not sure. I needed something for the HalRemote, but honestly at the time I wrote this I had no idea we'd have a full URL as a link on the story distro.

cavis:
This is getting close.  Still need to track down 1 bug on the initial story-creation (which isn't giving me any sane trace in our source code).

Oh, and fix the tests.

src/app/shared/model/distribution.model.ts
```diff
@@ -88,29 +100,21 @@ export class DistributionModel extends BaseModel {
...
+  // swap new episode with this.url
   saveRelated(): Observable<boolean[]> {
-    if (this.podcast && this.podcast.isNew) {
+    if (this.podcast && this.podcast.isNew && this.url) {
       let oldModel = this.podcast;
-
-      // CMS should actually have created the podcast in feeder
-      return this.loadExternal().flatMap(() => {
-        oldModel.copyTo(this.podcast);
+      this.loadRelated('podcast', true).subscribe((newModel: FeederPodcastModel) => {
```
sandikbarr: throws error because this isn't returning anything to map in BaseModel.save()

cavis: Yeah, seems that CMS isn't returning the story_distribution right away, after it's automagically created along with the story (there's now an open issue for that).

I'm currently refactoring parts of the BaseModel to allow for these records that are automagically created (the FeederPodcast, StoryDistribution, and FeederEpisode are all created on the backend).  Basically, after the parent model finishes saving and goes to `saveRelated()`, it will check for related that need to be reloaded first.

Getting close(r)!

src/app/series/directives/series-podcast.component.html
```diff
@@ -69,6 +69,10 @@
       <div class="fancy-hint">Use this to manually change the URL where your podcast is located.</div>
     </publish-fancy-field>
 
+    <publish-fancy-field label="Podcast Audio" [select]="audioVersionOptions" [model]="distribution" name="versionTemplateUrl" required>
```
cavis: Added a dropdown to allow picking which `set_audio_version_template_uri` to use

src/app/series/directives/series-podcast.component.ts
```diff
@@ -66,27 +55,35 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
...
-      this.templateSub.unsubscribe();
-    }
   }
 
-  loadPodcast() {
-    this.podcast = null;
-    if (this.distribution) {
-      this.distribution.loadExternal().subscribe(() => this.setPodcast);
+  loadDistributions() {
+    let initial = this.initialDistributions;
```
cavis: Was having some issues with change detection here.  So ended up recording what the initial `series.distributions` array looked like, and manually triggering changes when it is different.

src/app/shared/form/fancy-field.component.html
```diff
@@ -25,7 +25,7 @@ <h4 *ngIf="label && small">
       [(ngModel)]="model[name]" (ngModelChange)="onChange($event)"></textarea>
     <select *ngIf="type == 'select'" [id]="name"
       [(ngModel)]="model[name]" (ngModelChange)="onChange($event)">
-      <option *ngFor="let opt of select" [value]="opt">{{opt}}</option>
+      <option *ngFor="let opt of select" [value]="opt[1]">{{opt[0]}}</option>
```
cavis: Can now have display text that differs from the value now.

src/app/shared/model/base.model.ts
```diff
@@ -126,13 +128,34 @@ export abstract class BaseModel {
...
-      return this.relatedReplays[relName];
+      return this.relatedReplays[relName].first();
     } else {
-      let allRelated = this.RELATIONS.map(r => this.loadRelated(r, force).first());
+      let allRelated = this.RELATIONS.map(r => this.loadRelated(r, force));
       return Observable.forkJoin(allRelated).map(relateds => null);
     }
   }
 
+  swapRelated(): Observable<boolean[]> {
```
cavis: This is the black magic.  If a model defines a `swapNew()` method (couldn't figure out a very typescript-y way of checking for this, other than `if (model['swapNew'])`...

After saving itself, the parent model will force reloading that relation with `this.loadRelated('things', true)`.  Then it will swap out the previous `this.things` with the newly loaded values.  And allow the old model to copy/set its properties to the new model.

This is all to allow the user to set distribution/episode form data ahead of time, even though those records are actually created on the server side.

src/app/shared/model/base.model.ts
```diff
@@ -126,13 +128,34 @@ export abstract class BaseModel {
         });
         this.relatedReplays[relName] = relValue;
       }
-      return this.relatedReplays[relName];
+      return this.relatedReplays[relName].first();
```
cavis: This was driving me nuts.  Always call `first()` here so that the observable completes, and the `forkJoin` methods don't just hang forever.

src/app/shared/model/distribution.model.ts
```diff
@@ -40,46 +42,62 @@ export class DistributionModel extends BaseModel {
...
+    return {podcast, versionTemplate};
   }
 
   decode() {
     this.id = this.doc['id'];
     this.kind = this.doc['kind'] || '';
+
+    // TODO: cms has non-auth'd urls
     this.url = this.doc['url'] || '';
+    if (this.url && !this.url.match('/authorization/')) {
```
cavis: Feeder was returning 404s when the story was unpublished, so i switched to the `/authorized` url for the story.  I think there's a fix out so Feeder will actually redirect these requests.  But I'll check that later.

src/app/shared/model/story-distribution.model.ts
```diff
@@ -0,0 +1,74 @@
...
+    data.kind = this.kind;
+    return data;
+  }
+
+  saveNew(data: {}): Observable<HalDoc> {
+    throw new Error('Cannot directly create a story distribution');
+  }
+
+  swapNew(newModel: StoryDistributionModel) {
+    newModel.episode = this.episode;
```
cavis: This is a bit funky.  When you save the StoryModel, CMS is going to create both the StoryDistribution and the FeederEpisode on the backend.  So both the StoryDistributionModel and the FeederEpisodeModel are reloaded/swapped in instead of POSTing anything new.

src/app/story/directives/podcast.component.ts
```diff
@@ -1,96 +1,86 @@
 import { Component, OnDestroy } from '@angular/core';
```
sandikbarr: So close! This should also 
 - [ ] default the author name and email to the series author name and email
 - [ ] and show the enclosure url (readonly)

sandikbarr:
Looks good! Thanks @cavis!

PR #192, opened 2017-01-09
Allow user to set author email for series

### LINKED ISSUE #188: Author name is not prefilled on newly created Podcast Distribution
The author name should be prefilled with the account name for newly created podcast distributions. The default currently works for existing podcast distributions but not after just having clicked the Create Podcast button.

gcampo88:
Completes #167 and fixes #188


src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -127,8 +133,8 @@ export class FeederPodcastModel extends BaseModel {
     }
   }
 
-  set(field: string, value: any) {
-    super.set(field, value);
+  set(field: string, value: any, forceOriginal = false) {
```
sandikbarr: This function got removed from underneath you. Probably should do a merge from master and make sure it's goes away in this branch.

Issue #169, opened 2016-12-21
Story, Embeddable player staging host


kookster:
Use `play-staging.prx.tech` url on `publish-staging.prx.tech`

cavis:
@gcampo88 updated reqs: to get this also working in prod, we should add a `PLAY_HOST` to `env-example`.  Can default that to play-staging, since we tend to use the staging env for dev.

Will also need to update the staging and production ECS task definitions, to correctly set this config.

PR #198, opened 2017-01-10
Use env var for play host

gcampo88:
Use an environment variable for PLAY_HOST instead of putting that in the component; and update the auth host for staging. Finishes up #169.

### DIFF
--- a/src/app/story/directives/player.component.ts
+++ b/src/app/story/directives/player.component.ts
@@ -2,6 +2,7 @@ import { Component, OnDestroy, ElementRef, ViewChild } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { StoryModel, TabService } from '../../shared';
 import { DomSanitizer } from '@angular/platform-browser';
+import { Env } from '../../core/core.env';
 
 @Component({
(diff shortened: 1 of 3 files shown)


src/app/story/directives/player.component.ts
```diff
@@ -37,7 +37,7 @@ export class PlayerComponent implements OnDestroy {
   private audioUrl: string;
   private imageUrl: string;
   private subscriptionUrl: string;
-  private host: string = 'http://play-staging.prx.tech';
+  private host: string = window['ENV']['PLAY_HOST'];
```
cavis: Okay, so I may have once more misled you with my description in the ticket.  This `PLAY_HOST` should also be added to `core.env.ts`.  Then you `import { Env } from '../core.env'` into this file, and access it at `Env.PLAY_HOST`.

Looks like you should probably also set a DEFAULTS for this to `https://play.prx.org`.

Bleh - and reading that Env code for the first time in awhile, I realize that I don't really like it very much.  But it'll work for now!

gcampo88:
gotcha! back to you @cavis

Issue #257, opened 2017-01-17
Series, Podcast distribution, removing template


kookster:
Doesn't highlight when it is selected by tabbing - remove from tab order
Clicking/return on Delete link for remove template doesn't prompt

gcampo88:
I think we decided in conversation that taking the remove buttons out of the tab index (and thus largely removing the odds that someone will press them by accident) was enough, so that's what I'm doing. If we also want this to open up the confirmation modal we can reopen this. Ca va?

Issue #252, opened 2017-01-17
Story, embed player instructions are unclear


kookster:
don't use the term "clip", and change "story" to "episode" when part of a series (pending related ticket on story vs. episode).

sandikbarr:
Story has been changed to episode with PR #269. Adding to Jasmine and putting in Done.

Issue #251, opened 2017-01-17
Story, podcast distribution, release date hint could be clearer


kookster:
Indicate how this controls a future publish date, and how it can also be used to set a past date to fix ordering when putting in an older episode.

User questions:
"Does this mean I can set when the story will publish in the future?"
"Maybe something like "If you want to release at a later date or set to an earlier date, otherwise we'll publish immediately"?"

"Release Date isn't clear that it is also time, time wasn't clear - fix instructions and title, or somehow make clearer"

Issue #241, opened 2017-01-17
Series, Podcast distribution, homepage url scheme required is confusing


kookster:
On homepage url, need to tell people it needs "http", most people won't know to do that.

Instead handle it being missing, add hints/docs and error message that include this, or find other way to ask this (scheme select).

sandikbarr:
When adding a link in the markdown editor, if http:// or https:// is not provided, it prepends http://. However we handle it, we should make these url fields behave consistently.

kookster:
do the magic!

sandikbarr:
Too quick on the draw. If I start typing ht, I get "http://ht"

PR #269, opened 2017-01-17
Hint fixes, word choice changes, etc

gcampo88:
- [x] removes Add/Remove Segment + Template buttons from tab index on Series Audio Versions #257
- [x] changes 'clip' to 'story' on Story Player #252
- [x] clarifies Story Basic - Release Date hint #251
- [x] magically prepends `http://` onto Series Podcast Dist user-generated links if they are missing uri scheme #241
- [x] changes "story" to "episode" everywhere that user will see #235
- [x] changes phrasing on Series Audio Templates page #233


src/app/series/directives/series-templates.component.ts
```diff
@@ -12,39 +12,39 @@ import {
...
         <p>
-          When adding stories to your series, you can have different versions of the audio for
+          When adding episodes to your series, you can have different versions of the audio for
           distribution. For example, it is common to have a podcast version of the audio broken
           into segments for ad injection, but also a broadcast version with different breaks for
           local news, different credits, or edits and bleeps to meet FCC permitted language requirements.
-          On this page, you can define the types of audio versions that will be expected for each story in this
-          series, and basic requirements for each of those versions.
+          On this page, you can define basic templates for the versions of audio that will be expected for
+          each episode in this series, and requirements for each of those templates.
```
sandikbarr: minor: the comma between "series, and" seems off. I can see why there is a separation of those ideas, so maybe it would read better with the comma if it said "series, as well as". Thought it worth mentioning, but I'll defer to you on what correct grammar would be.

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -20,8 +20,7 @@ export class FeederPodcastModel extends BaseModel {
   authorEmail: string = '';
 
   VALIDATORS = {
-    link: [REQUIRED(), URL('Not a valid URL')],
-    newFeedUrl: [URL('Not a valid URL')]
```
sandikbarr: I think that we still need the URL validators even with the magic. (The Wysiwyg editor doesn't have good validation on the URL, which could be added in a separate ticket.) If we don't keep the validation here though, it will allow creation of urls such as "http://s". Even though Ryan says it's "not entirely accurate" other URL validation I've worked with is incredibly loose, mostly just requiring a period with valid characters something before and after. So what we have is better than not having it I think, might have to fudge with the 'http|https' part of the test though depending on what string it is validating.

cavis: I'd agree that allowing someone to save a non-valid URL is a bad idea.  Saving a blank may be okay.

gcampo88: :+1:

src/app/series/directives/series-templates.component.ts
gcampo88: 👍

gcampo88:
good call @sandikbarr, back to you with changes!

PR #271, opened 2017-01-17
Use feeder for embed player

cavis:
Load embed player from different sources:

- [x] For published stories in feeder, use the feed-url + guid for the player
- [x] For unpublished stories that will eventually be in feeder, preview the player via query params, but still use the url/guid for the copyable link/iframe.  Displays an angry warning about what you're doing.
- [x] For stories not in feeder, just keep using the query params
- [x] Handle various distribution/podcast/episode load errors

Note that any changes to CMS will take a few seconds to propagate to feeder.  So it's fairly easy to get a broken embed player, if say, you just clicked the Publish button.

There are a couple feeder issues that will make this look not-entirely convincing at the moment, including: PRX/feeder.prx.org#177 and PRX/feeder.prx.org#181.

### DIFF
--- a/src/app/story/directives/player.component.ts
+++ b/src/app/story/directives/player.component.ts
@@ -1,122 +1,157 @@
-import { Component, OnDestroy, ElementRef, ViewChild } from '@angular/core';
+import { Component, OnDestroy, DoCheck } from '@angular/core';
 import { Subscription } from 'rxjs';
-import { StoryModel, TabService } from '../../shared';
-import { DomSanitizer } from '@angular/platform-browser';
+import { StoryModel, DistributionModel, TabService } from '../../shared';
+import { DomSanitizer, SafeResourceUrl } from '@angular/platform-browser';
 import { Env } from '../../core/core.env';
+
+const DEFAULT_IMAGE = 'http://s3.amazonaws.com/production.mediajoint.prx.org/public/comatose_files/4625/prx-logo_large.png';
+const DEFAULT_AUDIO = 'http://s3.amazonaws.com/production.mediajoint.prx.org/public/comatose_files/0000/meet-prx.mp3';
 
 @Component({
   styleUrls: ['player.component.css'],
-  template: `
-    <div *ngIf="story">
-      <p>
-        You can include this embeddable player on your website to play the audio for this clip directly.
-        Here's a preview of what the player will look like for this story. Note that the image and audio
-        are placeholders until this story has been published.
-      </p>
-        <iframe
-                [src]='this.safeUrl'
-                frameborder="0"
-                height="200"
-                width ="100%">
-        </iframe>
-
-        <p>You can embed this player on your own site by including the following <code>iframe</code> tag.</p>
-
-        <div class="embed-code">
-          <input type="text" [value]="iframeHtml(200, 650)" #shareEmbedSmall readonly>
-          <button (click)="copy(small)" data-copytarget="#shareEmbedSmall" #small>Copy</button>
-        </div>
-
-    </div>
-  `
+  templateUrl: 'player.component.html'
 })
 
-export class PlayerComponent implements OnDestroy {
-  @ViewChild('shareEmbedSmall') private inputEl: ElementRef;
+export class PlayerComponent implements OnDestroy, DoCheck {
+
+  private tabSub: Subscription;
+  private story: StoryModel;
+
+  private feedUrl: string;
+  private episodeGuid: string;
   private title: string;
   private subtitle: string;
   private audioUrl: string;
   private imageUrl: string;
   private subscriptionUrl: string;
 
-  story: StoryModel;
-  tabSub: Subscription;
+  shouldUseFeeder: boolean;
+  previewingUnpublished: boolean;
+  loadError: string;
 
-  constructor(
-    tab: TabService,
-    private sanitizer: DomSanitizer
-  ) {
-    this.tabSub = tab.model.subscribe((s: StoryModel) => {
-      this.setFields(s);
+  previewUrl: string;
+  previewSafe: SafeResourceUrl;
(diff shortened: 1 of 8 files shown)


cavis:
@gcampo88 giving you this review, since you wrote this original component!  May not be possible to entirely QA all these features given those 2 feeder bugs.

src/app/story/directives/player.component.ts
```diff
@@ -1,122 +1,157 @@
...
 
-    str.push(`tt=${this.encode(this.title)}`);
-    str.push(`ts=${this.encode(this.subtitle)}`);
-    str.push(`ua=${this.encode(this.audioUrl)}`);
-    str.push(`ui=${this.encode(this.imageUrl)}`);
-    str.push(`us=${this.encode(this.subscriptionUrl)}`);
-    return str.join('&');
-  }
+      // some defaults in case we can't find data
+      this.subscriptionUrl = 'http://www.prx.org';
```
cavis: The player now requires some "minimum" fields in order to render at all (title + subtitle + image + audio + subscribe).  Was trying to make sure those got set here no matter what.

src/app/story/directives/player.component.ts
```diff
@@ -1,122 +1,157 @@
...
+    let feeder: string[] = [];
+    feeder.push(`uf=${this.encode(this.feedUrl)}`);
+    feeder.push(`ge=${this.encode(this.episodeGuid)}`);
+    let feederUrl = `${Env.PLAY_HOST}/e?${feeder.join('&')}`;
+
+    // only change the safe resource when the url actually changes
+    let newPreview = previewFeeder ? feederUrl : cmsUrl;
+    if (this.previewUrl !== newPreview) {
+      this.previewUrl = newPreview;
+      this.previewSafe = this.sanitizer.bypassSecurityTrustResourceUrl(newPreview);
```
cavis: Turns out, every time you get `bypassSecurityTrustResourceUrl`, you get a new/unique value.  So calling this every time causes the iframe to flicker like crazy due to change detection.

src/app/story/directives/player.component.ts
gcampo88: If user has no uploaded audio or image though, they will still see the preview of the player, with the default audio and default image (and no preview message). Seems a little funky to not say anything about that. Maybe we could show them a message saying here's what the player will look like, for now it's using our default image and audio but once you provide those items it'll use yours - or something?

cavis: Hmm... yeah, that makes sense.  Good sense!

The non-feeder player is also a bit of a TODO... since it doesn't let you select which audio-version or image to use.  And it can't play > 1 mp3 file in an audio-version.  But i figured those edge cases are pretty low priority.

gcampo88: 👍

cavis: Also ... I may have forgotten to ever use a non-default subscription url.  (Should probably be the beta.prx.org story URL).

gcampo88: nah, we can just make everyone subscribe to www.prx.org

gcampo88:
Some issues when I publish a valid story within a podcast, and am looking at the player page: 
1) In the first 10 seconds or so after clicking Publish, I get this in the console: `feed.adapter.ts:14 Could not find item with guid prx:90:7563ee2d-0e73-4a16-b17e-3edfe5ecfc48` and player just looks broken-- I assume that's because of the delay from when something is published to when it's available in feeder. Maybe we could set a timeout or delay the time that Story.isPublished checks against by 30 seconds or so, so the player doesn't try to pull from feeder before the episode is there? and/or show a message to user explaining the delay and why the player looks broken for the moment? 
2) After that, the "Could not find item with guid ..." error goes away, but the player still looks broken -- no audio, no image, and the only error I can get is `Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().` when I click play and then pause (on the empty audio). Is that because of #177?

src/app/story/directives/player.component.ts
gcampo88: good catch!

cavis:
Per (1) ... yeah, I'll put in some sort of a delay option on `isPublished()`, and display a warning or something the first N seconds after the story is published.

Per (2) ... yes, that's Feeder returning broken image/audio file links.

gcampo88:
Sounds good. Looks like you may also need a merge master, and then good to go with those changes!

cavis:
@gcampo88 still a bit tricky to test, given the feeder-410s, broken links, and http-vs-https errors.  But I think i've addressed the listed issues.

src/app/story/directives/player.component.ts
```diff
@@ -1,122 +1,175 @@
...
 
-    if (this.story.versions.length > 0) {
-      let firstAudio = this.story.versions[0].files.find(file => !file.isDestroy);
-      if (firstAudio) { this.audioUrl = firstAudio.enclosureHref; }
+  ngDoCheck() {
+    let hasCmsParams = this.title && this.subtitle && this.audioUrl && this.imageUrl && this.subscriptionUrl;
+    let hasFeederParams = this.feedUrl && this.episodeGuid;
+    if (this.shouldUseFeeder === false && hasCmsParams) {
+      this.previewingUnpublished = false;
+      this.setCmsLoadError();
```
gcampo88: Is it right that you only see the message about default audio + image being in the preview if this is a standalone story? Since cover image isn't required in Basic tab, very possible that a user creating a story in a podcast will get here without an imageUrl, in which case the unpublished preview they see will have the default image. Fine to leave that as is? I've argued both sides to myself. 
In any case, I'll approve as is and you can do as you see fit!

cavis:
Fallout: #278.  Onwards!

Issue #256, opened 2017-01-17
Confirm critical field change after edit


kookster:
For some fields, like 'new feed url' on a podcast, changing them can have dramatic effect.
For such fields, have on focus change after edit cause a confirm message to appear warning the user of the effect, and giving them a chance to undo the field change, or continue

Issue #240, opened 2017-01-17
Series, Podcast Distribution, New feed url should be moved and explained


kookster:
New Feed URL should have a more significant warning (not on this page maybe, or in an advanced, or separated) - R

New feed url - I don't know what that means. I know b/c of Robert, but otherwise I wouldn't know - a new person would not. - G

kookster:
https://help.apple.com/itc/podcasts_connect/#/itcb54353390
"Use the <itunes:new-feed-url> tag to manually change the URL where your podcast is located. This tag is only supported at a <channel>level (podcast).

<itunes:new-feed-url>http://newlocation.com/example.rss</itunes:new-feed-url>
Note: You should maintain your old feed until you have migrated your existing subscribers. For more information, see Update your RSS feed URL."

PR #272, opened 2017-01-18
advanced field confirmation for new feed url

sandikbarr:
#256 and #240 
If the new feed url is changed to a value that passes validation on a not new podcast, the user will be prompted to confirm the change. If the user clicks Cancel, the change is reverted.

### DIFF
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -65,12 +65,15 @@
       <div class="fancy-hint">A link to the homepage for your podcast.</div>
     </publish-fancy-field>
 
-    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl">
-      <div class="fancy-hint">Use this to manually change the URL where your podcast is located.</div>
-    </publish-fancy-field>
-
     <publish-fancy-field label="Podcast Audio" [select]="audioVersionOptions" [model]="distribution" name="versionTemplateUrl" required>
       <div class="fancy-hint">Select which audio version template should be used with your podcast.</div>
+    </publish-fancy-field>
+
+    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl" [advanced]="newFeedUrlConfirm">
+      <div class="fancy-hint">
+        If your podcast feed is moving, use this field to point to the new URL where your podcast is located.
+        The current feed should be maintained until all of your subscribers have migrated.
+      </div>
     </publish-fancy-field>
 
   </div>
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -69,7 +69,7 @@
       <div class="fancy-hint">Select which audio version template should be used with your podcast.</div>
     </publish-fancy-field>
 
-    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl" [advanced]="newFeedUrlConfirm">
+    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl" [advancedConfirm]="newFeedUrlConfirm">
       <div class="fancy-hint">
         If your podcast feed is moving, use this field to point to the new URL where your podcast is located.
         The current feed should be maintained until all of your subscribers have migrated.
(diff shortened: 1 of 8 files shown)


src/app/shared/form/fancy-field.component.ts
```diff
@@ -1,5 +1,6 @@
 import { Component, Input, Output, EventEmitter } from '@angular/core';
 import { BaseModel } from '../model/base.model';
+import { AdvancedConfirm } from './advanced-confirm.directive';
```
cavis: Seeing this warning from webpack on npm start ...

```
WARNING in ./src/app/shared/form/fancy-field.component.ts
174:55 export 'AdvancedConfirm' was not found in './advanced-confirm.directive'

WARNING in ./src/app/shared/form/fancy-field.component.ts
174:90 export 'AdvancedConfirm' was not found in './advanced-confirm.directive'
```

src/app/series/directives/series-podcast.component.html
```diff
@@ -65,14 +65,17 @@
...
 
-    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl">
-      <div class="fancy-hint">Use this to manually change the URL where your podcast is located.</div>
-    </publish-fancy-field>
-
     <publish-fancy-field label="Podcast Audio" [select]="audioVersionOptions" [model]="distribution" name="versionTemplateUrl" required>
       <div class="fancy-hint">Select which audio version template should be used with your podcast.</div>
     </publish-fancy-field>
 
+    <publish-fancy-field label="New Feed Url" textinput [model]="podcast" name="newFeedUrl" [advanced]="newFeedUrlConfirm">
```
cavis: Since we're apparently going to be adding this to many fields, i wonder if this interface can be simplified a bit.

The fancy-field already has the model and field-name, so all it really needs is the warning message.  Something like `[confirm]=confirmNewFeedUrlMsg`.  I'm assuming the message string is long enough that it's not worth cramming directly into the element attribute.

Or you might be able to something tricky with a child `<div class="fancy-confirm">Are you sure?</div>`.  But I'm not confident that's actually possible.

src/app/series/directives/series-podcast.component.ts
```diff
@@ -101,4 +101,21 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
...
+    if (!confirm) {
+      this.podcast.set('newFeedUrl', this.podcast.original['newFeedUrl']);
+    }
+  }
+
+  get newFeedUrlConfirm() {
+    if (this.podcast) {
+      return {
+        confirm: AdvancedConfirmText.newFeedUrl(
+          !this.podcast.isNew && !this.podcast.invalid('newFeedUrl') && this.podcast.changed('newFeedUrl'),
```
cavis: I have a feeling that checking for `invalid()` may not work out, in the future.  Assuming we allow saving invalid stuff to CMS.  But since this is a series and not a story ... I haven't quite hashed that out in my brain yet.

src/app/shared/form/advanced-confirm.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+  newFeedUrl: function (changed, originalValue, changedValue) {
+    if (changed) {
+      let prompt = "Are you sure you want to change New Feed URL";
+      if (originalValue) {
+        prompt += ` from "${originalValue}"`;
+      }
+      if (changedValue) {
+        prompt += ` to "${changedValue}"`;
+      }
+      prompt += '? This will point your subscribers to a new feed location.';
```
cavis: If you did just pass in the entire string via the `[confirm]="someConfirmationMessage"`, this directive could be much more field-agnostic.

src/app/series/directives/series-podcast.component.ts
```diff
@@ -101,4 +101,21 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
     }
   }
 
+  resetNewFeedUrlOnCancel(confirm) {
```
cavis: At first I found the "Cancel == revert change" thing a bit confusing.  But calling that button "Discard" would be even more confusing, since it just discards the one field, not the whole form.  So I think I'm good with this.

sandikbarr:
@cavis I made the advanced confirm directive model aware. The callback is within the directive and will now apply the same !isNew && !invalid && changed test wherever it's used. Hoping this works better for your purpose.

cavis:
👍  looks good to me.

Issue #275, opened 2017-01-18
Change how validations work with publishing and saving


kookster:
@kookster commented on [Tue Jan 17 2017](https://github.com/PRX/meta.prx.org/issues/21)

tbd.

---

@cavis commented on [Wed Jan 18 2017](https://github.com/PRX/meta.prx.org/issues/21#issuecomment-273531554)

Initial thoughts:

## Series

- Generally, a Series is considered "always published".  So you cannot ever save invalid/in-progress data.  Validations must pass, even on a new Series.
- But there aren't too many required fields.  So this won't really be an issue.
- In other words, Series doesn't change from how it is today, except that you can't navigate away without saving or discarding.

## Story

- New: there are some very basic requirements to create a story in CMS.  (Non-blank title is probably the only one).
- Unpublished: all validations displayed (so you know your TODOs), but only some prevent you from saving.
  - Fields are generally allowed blank, but must validate if present.
  - Similarly, audio/images are allowed blank, but must validate if present.
  - Exception: don't enforce the audio-version-template length validations, as they would prevent you from saving only 1-of-many audio-files.
  - Publish button only enabled when ALL validations pass
- Published: all validations displayed and enforced.
  - But if you unpublish at this point, you're back to only partial enforcement.

PR #281, opened 2017-01-19
Allow saving invalid unpublished stories

cavis:
Loosens up story validations per #275.

- New or Unpublished stories can save with only a title.
  - Other fields/files/etc aren't required.  But if you enter a non-blank thing, it's going to be validated
- Publish button is disabled until strict validations pass
- Published stories disable the save button until strict validations pass

Also cleaned up the dashboard/search cards a bit, to make them load less data.  And to standardize on what sort of statuses we display in them.

### DIFF
--- a/src/app/story/directives/hero.component.ts
+++ b/src/app/story/directives/hero.component.ts
@@ -29,7 +29,7 @@ import { StoryModel } from '../../shared';
           (click)="discard()">Discard</publish-button>
 
         <template [ngIf]="stateNew">
-          <publish-button green [model]="story" [disabled]="normalInvalid"
+          <publish-button green=1 [model]="story" [disabled]="normalInvalid"
               visible=1 (click)="save()">Create
             <div *ngIf="normalInvalid" class="invalid-tip create">
               <h4>Invalid episode</h4>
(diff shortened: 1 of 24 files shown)


src/app/story/directives/hero.component.ts
```diff
@@ -75,7 +96,13 @@ export class StoryHeroComponent implements OnInit, OnChanges {
   }
 
   ngDoCheck() {
-    this.now = new Date(); // expression changed after checked if new Date() in getter, but it does need to get updated on change cycles
```
sandikbarr: glad you got rid of this nonsense! thank you

sandikbarr:
Not entirely sure why this build is failing yet, but otherwise this should be ready to merge.

cavis:
Turned out to be the `Intl` api not being supported in Phantom.  I just stubbed out the `date` pipe for now.  Looks like there's a [bigger discussion](https://github.com/angular/angular/issues/10809) happening at angular around this.

Agreed on the "Draft" indicator - was trying to decide if "Draft" or "Unpublished" would be less controversial.  So I put the more controversial one in there to solicit feedback!

Issue #268, opened 2017-01-17
Story edit, unsaved changes warning says "save" instead of "create"


kookster:
We are talking about reworking this text anyway, but whatever it says should match actions available.

"Why do I have to cancel and then hit save on "Unsaved Changes" modal?
Says hit "save" but it says "create" on the page"

Issue #253, opened 2017-01-17
Navigating w/o saving and locally saved changes confusing


kookster:
Unsaved changes - that last sentence, I'm not sure what that means, stored in my browser? but how do I get back there?

Why would I want to not save my changes? Can't see why I wouldn't want to do that. Never something I wanted to do - I save or I don't.

To me draft means I haven't published yet.
(Seems like "draft" should only happen we have an unsaved doc we rescued from being lost, and we can come up with a different name for that)

sandikbarr:
The user will have to either Save or Discard their changes before navigating away. (Rules for allowing saving of incomplete stories will be adjusted with #275). The modal will have two buttons, Okay and Cancel. Cancel closes the modal and allows the user to review the changes made and either Save or Discard. Okay will discard changes and continue navigation.

PR #282, opened 2017-01-20
Unsaved changes forces save or discard

sandikbarr:
#253, #261, #268 which are all closely related, so one PR to force user to save or discard their changes to stories and series

### DIFF
--- a/src/app/core/modal/modal.component.css
+++ b/src/app/core/modal/modal.component.css
@@ -23,7 +23,7 @@
   max-height: 100%;
   width: 500px;
   margin-left: -250px;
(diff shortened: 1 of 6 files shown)


src/app/core/modal/modal.component.css
```diff
@@ -23,7 +23,7 @@
   max-height: 100%;
   width: 500px;
   margin-left: -250px;
-  height: 200px;
+  min-height: 200px;
```
sandikbarr: #261 to let the modal grow to the size of its content (but not larger than 100%)

Issue #266, opened 2017-01-17
Related field validations (min & max seconds) triggering before editing complete


kookster:
After editing min, get an error about it being greater than max, before max has even been edited.
Not sure how to solve this - delay on the validation until after max has lost focus or 30 seconds later, whichever comes first? Or just have it when min loses focus instead of on changing the value?
Or we could also do nothing - this is not so bad on its own, but I think was exacerbated by the general difficulty of editing the min and max fields. We could just fix that editing in general, and leave the validations as is for the next round of testing.

"Why is it chiding me about the min being less than max when max isn't set yet?"

cavis:
Decided to make the max fields "chase" the min.  So if you set min >= max, it sets `max = min + 60`.  This way, you never get a validation error while setting the min, which usually happens first.

You can still get a validation error by explicitly setting the max < min.

Issue #237, opened 2017-01-17
Series, Templates, fill out min max time in H:M:S not just total seconds


kookster:
I don't think most people have handy the duration in seconds, so would be easier to do this in minutes, but seconds would be tough.

"Thought the min and max on the audio template was for minutes not seconds"

cavis:
Went with an HH:MM:SS series of 3 tiny number inputs.  Actually works pretty well, since I can roll over values > 60 into the next higher field using angular magic.

PR #287, opened 2017-01-23
Better series templates

### LINKED ISSUE #236: Series, Templates, Explain why min and max are needed
Not clear why these need to be filled out or how they will be used.

cavis:
Fixes #236 #237 #266.

- [x] Update the series-templates tab UI to look more like the story audio-templates
  - Ended up doing some "funny" stuff with the forms, to make this all fit.  Let me know what you think.
- [x] Use an HH:MM:SS picker for the min/max length fields, instead of just seconds
- [x] Avoid showing min/max length validation errors too early
  - Aggressively keeps the max above the min, when editing the min.  But you can get validation errors when setting the max.  Let me know what you think - I didn't have any better ideas that didn't break our "show validation errors immediately" paradigm.
- [x] Try to fix some bugs when you discard on a new series, and the default "Podcast" template disappears.
  - Getting better here, but adding the default audio version/file templates to the series is still a bit funky.
- [x] Wordsmithing, but probably not very good wordsmithing.


src/app/shared/form/fancy-duration.component.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { cit, create, stubPipe } from '../../../testing';
+import { FancyDurationComponent } from './fancy-duration.component';
+
+fdescribe('FancyDurationComponent', () => {
```
gcampo88: fdescribe!

src/app/series/directives/series-templates.component.ts
```diff
@@ -11,59 +11,52 @@ import {
...
             </publish-fancy-field>
-            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMaximum" label="Maximum">
+
+            <publish-fancy-field class="length" [model]="v" label="Total length" invalid="lengthAny">
+              <div class="fancy-hint">
+                The minimum and maximum durations for all the audio files. Used to check
+                that your episodes have sane lengths, and prevent uploading bad audio.
+              </div>
+              <publish-fancy-duration [model]="v" name="lengthMinimum" label="Minimum"></publish-fancy-duration>
+              <publish-fancy-duration [model]="v" name="lengthMaximum" label="Maximum"></publish-fancy-duration>
```
gcampo88: possible to make the hh/mm/ss inputs line up vertically when screen is small enough to make them stack?
screen shot 2017-01-24 at 11 19 59 am

src/app/series/directives/series-templates.component.ts
```diff
@@ -11,59 +11,52 @@ import {
...
+          <section>
+            <publish-fancy-field required textinput [model]="v" name="label" label="Template Label">
+              <div class="fancy-hint">A name for this audio template, such as "Podcast Audio" or "Clean Version"</div>
             </publish-fancy-field>
-            <publish-fancy-field number small inline hideinvalid [model]="v" name="lengthMaximum" label="Maximum">
+
+            <publish-fancy-field class="length" [model]="v" label="Total length" invalid="lengthAny">
+              <div class="fancy-hint">
+                The minimum and maximum durations for all the audio files. Used to check
+                that your episodes have sane lengths, and prevent uploading bad audio.
```
gcampo88: I'd vote against "sane" -- sounds like we'd be judging their episode lengths by standards other than what they're setting here. Maybe something like "Used to ensure that each of your episodes is the desired approximate length, and to prevent uploading bad audio."

src/app/series/directives/file-template.component.ts
```diff
@@ -1,33 +1,54 @@
...
+        <publish-fancy-field textinput required hideinvalid [model]="file" name="label">
         </publish-fancy-field>
+      </div>
 
-        <publish-fancy-field number small inline hideinvalid [model]="file" name="lengthMaximum" label="Maximum">
-        </publish-fancy-field>
+      <div class="lengths">
+        <publish-fancy-duration [model]="file" tiny="true" name="lengthMinimum" label="Min"></publish-fancy-duration>
+        <publish-fancy-duration [model]="file" tiny="true" name="lengthMaximum" label="Max"></publish-fancy-duration>
+      </div>
```
gcampo88: I love the one-liner matching the story audio uploader -- I think it makes it make much more intuitive sense to user on both pages. But the styling starts to look weird as it wraps onto next line as screen shrinks: 

screen shot 2017-01-24 at 11 25 42 am

maybe this just shouldn't wrap?

src/app/shared/form/fancy-duration.component.spec.ts
cavis: dangit!  no wonder my tests passed so easily.

src/app/series/directives/file-template.component.ts
cavis: Hmm... yeah, i'll take a shot at making this at least somewhat responsive.  Probably only down to tablet-width, since Publish generally doesn't work so good when you get down to phone size.

src/app/series/directives/series-templates.component.ts
cavis: I was hoping someone would have a better idea!

sandikbarr: Yes, let's let the podcaster continue being their best crazy selves. We're just helping them set limits. 😄

src/app/shared/form/fancy-duration.component.ts
```diff
@@ -0,0 +1,83 @@
...
+@Component({
+  selector: 'publish-fancy-duration',
+  styleUrls: ['fancy-field.component.css', 'fancy-duration.component.css'],
+  template: `
+    <div class="field small inline" [class.tiny]="tiny">
+      <h4 *ngIf="label">
+        <label [attr.for]="hoursName">{{label}}</label>
+      </h4>
+      <template [ngIf]="!model">
+        <input [id]="hoursName" type="number" disabled=true/>
```
sandikbarr: I am disappointed in myself for not using input type number with step, min, and max on the time picker. I will enter a ticket.

cavis: Yeah, the up/down arrows are real handy, if you're tabbing through these fields.

src/app/shared/form/fancy-duration.component.ts
```diff
@@ -0,0 +1,83 @@
...
+        <b>:</b>
+        <input [id]="minutesName" type="number" disabled=true/>
+        <b>:</b>
+        <input [id]="secondsName" type="number" disabled=true/>
+      </template>
+      <template [ngIf]="model">
+        <input [id]="hoursName" type="number" min="0" [ngModel]="hours | padzero"
+          (ngModelChange)="set('hours', $event)" [class.changed]="hoursChanged"/>
+        <b>:</b>
```
cavis: I guess my reasoning was partially non-semantic, since I needed to style the `:` without changing the `<input>` font style as well, so it needed its own element.  Which also let me set specific margins on that single character, to space the inputs properly.

But if I pretend I was thinking semantically for a second... the `<b>` tag "represents a span of text stylistically different from normal text, without conveying any special importance or relevance".  Which is exactly what I'm after - stylistically different but not important.

src/app/series/directives/file-template.component.css
```diff
@@ -1,20 +1,58 @@
...
+.remove {
+  text-align: right;
+  order: 3;
+  flex-basis: 50px;
+  padding-right: 10px;
+}
+.remove i {
+  color: #368AA2;
+  font-size: 22px;
+  cursor: pointer;
```
cavis: I get way too in the habit of making "whatever the heck I want" clickable.  Buttons it is!

cavis:
@gcampo88 responsivized!

Going to change some `<i>`s to buttons, as that's how I'm treating them anyways.  But then I think this is about good.

cavis:
@sandikbarr changed to buttons.  And changed the story-audio-uploader icons to buttons as well, since I copied from there in the first place.  Should be ready to go if that didn't break the tests.

gcampo88:
:+1:

Issue #290, opened 2017-01-24
story, podcast distribution guid in advanced settings


sandikbarr:
The guid in the story podcast distribution should be in a hidden-by-default advanced settings section. ~If changed by the user, the advanced confirm modal should be shown further explaining to the user the implications of changing the item's guid.~

~Question: Not sure but it seems to make sense to confirm the change whether the story is published or not. It's possible that they previously had it published and should understand what it does to change the guid.~ covered in #277

PR #294, opened 2017-01-24
guid goes in advanced settings

sandikbarr:
#290


src/app/story/directives/podcast.component.ts
```diff
@@ -24,7 +24,6 @@ export class PodcastComponent implements OnDestroy {
 
   tabSub: Subscription;
   storyDistribution: StoryDistributionModel;
-  seriesDistribution: DistributionModel;
```
sandikbarr: I was chasing down errors from bad data and saw this variable is not used.

Issue #303, opened 2017-01-26
Add max-length validation to appropriate fields


gcampo88:

gcampo88:
- series title
- story title
- audio version template label
- audio version label
- audio file template label
- audio file label

Issue #302, opened 2017-01-25
Search Stories, sort orders


kookster:
- [x] For last updated at, add it to the story card or remove it from sort options
- [x] Sort by published at, if it is unpublished, leave the date blank instead of using updated_at
- [x] For now don't use the released_at instead of published_at and get user feedback on it

Issue #56, opened 2016-10-06
search story card should show play count


sandikbarr:
Currently the search story card is showing audio file count where it should be showing play count. The icon being used is also not the correct icon for play count.

PR #304, opened 2017-01-26
Update Story Search options and appearance

gcampo88:
#302: 
- [x] removes updated_at from Story search options and never displays it on story card
- [x] only shows `published_at` date on Story card (won't show a `released_at` for a saved but unpublished story)
- [x] reinstates Story's ability to filter by `published_at` by fixing typo
- [x] clarifies that the date you see on Series is indeed `updated_at`
- [x] refactors story card styling slightly

#303: 
- [x] adds max-length validations of 255 char to all fields that have column type string in CMS.

### DIFF
--- a/src/app/search/directives/story-card.component.css
+++ b/src/app/search/directives/story-card.component.css
@@ -69,14 +69,14 @@ publish-image  {
   display: -ms-flexbox;
   display: flex;
 }
-.duration, .modified, .play-count {
+.duration, .published, .play-count {
   width: calc(100%/3);
 }
 .play-count {
   flex: 2 2 60px;
   white-space: nowrap;
 }
-.duration, .modified {
+.duration, .published {
   flex: 1 1 60px;
 }
 .play-count > i {
--- a/src/app/search/directives/story-card.component.css
+++ b/src/app/search/directives/story-card.component.css
@@ -70,9 +70,6 @@ publish-image  {
   display: flex;
   justify-content: space-between;
 }
-.play-count > i {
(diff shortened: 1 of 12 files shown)


src/app/search/directives/story-card.component.css
```diff
@@ -68,16 +68,7 @@ publish-image  {
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;
-}
-.duration, .modified, .play-count {
-  width: calc(100%/3);
-}
-.play-count {
```
sandikbarr: This was hanging out for #56 but things will probably change by the time we have that info, so yeah no sense working around something we don't have yet.

src/app/search/directives/story-card.component.ts
```diff
@@ -18,7 +17,7 @@ import { StoryModel } from '../../shared';
       <section class="story-info">
         <span class="duration">{{storyDuration | duration}}</span>
         <span class="play-count"><i></i></span>
-        <span class="modified">{{storyDate | date:"MM/dd/yy"}}</span>
+        <span class="published">{{storyDate | date:"MM/dd/yy"}}</span>
```
sandikbarr: I don't see a published class related to this. Should remove class= duration and play-count too if getting rid of those.

gcampo88: good call! My oversight-- I had made rules for `.published` but then removed them when refactoring the CSS.

gcampo88:
good call. back to you @sandikbarr !

Issue #297, opened 2017-01-25
Update released at control


kookster:
- [x] Have a checkbox to say you want to release on a specific date (past or future), unchecking clears the date
- [x] Update the time control to be a single drop down with all the hour + half hour, am/pm values (~48)
- [x] Add timezone
- [x] Do not reset time to midnight when a new date is selected

PR #306, opened 2017-01-26
releasedAt updates

sandikbarr:
#297

### DIFF
--- a/src/app/shared/date/timepicker.component.ts
+++ b/src/app/shared/date/timepicker.component.ts
@@ -3,11 +3,10 @@ import { Component, Input, Output, EventEmitter } from '@angular/core';
 @Component({
   selector: 'publish-timepicker',
   template: `
-    <input [class.changed]="changed" type="number" min="0" max="24" 
-      [ngModel]="hour | padzero" (ngModelChange)="set('hour', $event)">
-    <span class="separator">:</span>
-    <input [class.changed]="changed" type="number" min="0" max="30" step="30"
-      [ngModel]="minutes | padzero" (ngModelChange)="set('minutes', $event)">
+    <select [class.changed]="changed"
+      [ngModel]="time" (ngModelChange)="set($event)">
+      <option *ngFor="let o of options" [value]="o">{{o}}</option>
+    </select>
   `,
   styleUrls: ['timepicker.component.css']
 })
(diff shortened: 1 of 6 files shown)


src/app/shared/date/timepicker.component.ts
```diff
@@ -19,40 +15,94 @@ export class TimepickerComponent {
...
-      let date = new Date(this.date.valueOf());
-      if (date.getMinutes() <= 30) {
-        return date.getHours() < 10 ? '0' + date.getHours() : '' + date.getHours();
-      } else {
-        return date.getHours() + 1 <= 23 ? '' + (date.getHours() + 1) : '00';
-      }
-    }
+
+  localTimezone = new Date().toString().match(/(\([A-Za-z\s].*\))/)[1];
+  options: string[] = [
```
cavis: Trying to think of a clever way to generate this.  Lots of crazy math solutions out there, but the one i like best just uses Date...

```
let times = [];
let day = new Date(1970, 0, 1, 0, 0, 0, 0);
while (day.getDate() === 1) {
  times.push(day.toTimeString());
  day.setMinutes(day.getMinutes() + 30);
}
```

sandikbarr:
@cavis @kookster @gcampo88 I'll reverse the checkbox since I'm 2 against 1 on that now. Are we thinking most users will not use this field and that's why default behavior is to disabled it? My thinking is that it will be frustrating for the user to try click a disabled datepicker and to have to discover (by reading, which users don't usually do) that they need to click a checkbox to enable it. Do we have a disabled field indicator? PRX Feed url comes to mind, looks same as other fields. At a minimum, I will use cursor: not-allowed on the disabled date and time pickers and should have on the disabled checkbox.

gcampo88:
What about not having the datepicker show up at all unless they click the checkbox?

cavis:
No strong opinion for me.  Just a lot of feelings.

I do _feel_ that using this field will not be the normal workflow.

src/app/shared/date/timepicker.component.ts
sandikbarr: @cavis Do you also prefer this toTimeString() format that includes timezone offset? i.e. `14:39:07 GMT-0600 (PDT)` instead of `14:39:07 (PDT)`

cavis: No - i like the way you had it.  You may have to do some string mutating.  If that `toTimeString()` output is guaranteed to have that format, you may be able to just regexp that?  Or assembling the pieces manually would work.

src/app/shared/date/timepicker.component.ts
```diff
@@ -17,56 +17,15 @@ export class TimepickerComponent {
...
-    '10:30pm ' + this.localTimezone,
-    '11:00pm ' + this.localTimezone,
-    '11:30pm ' + this.localTimezone
-  ];
+  options: string[] = [];
+
+  constructor() {
+    let day = new Date(1970, 0, 1, 0, 0, 0, 0);
+    while (day.getDate() === 1) {
+      this.options.push(this.dateToHumanTime(day));
```
cavis: 👍  Slick!

PR #308, opened 2017-01-26
Podcast template picker updates

### LINKED ISSUE #227: Podcast Distribution, setting the template
The select for audio template has no blank option, just the one version - should it default when there is only one version? I think it would be better to have a blank option as well as the one template.

cavis:
Fixes #227.

- If there's only 1 audio-version-template for the series, just default to that when creating the podcast
- If > 1, make them explicitly pick it
- Annoy-on-change for the dropdown
- Fix some linting with the advanced confirm directive (doesn't seem to like `@Input('name') differentName`)

### DIFF
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -49,7 +49,8 @@
       </div>
     </publish-fancy-field>
 
-    <publish-fancy-field label="Podcast Audio" [select]="audioVersionOptions" [model]="distribution" name="versionTemplateUrl" required>
+    <publish-fancy-field required label="Podcast Audio" name="versionTemplateUrl"
+      [select]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
       <div class="fancy-hint">Select which audio template should be used with your podcast.</div>
     </publish-fancy-field>
 
--- a/src/app/series/directives/series-podcast.component.ts
+++ b/src/app/series/directives/series-podcast.component.ts
@@ -17,6 +17,7 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   itunesExplicitDoc = 'https://support.apple.com/en-us/HT202005';
   itunesCategoryDoc = 'https://help.apple.com/itc/podcasts_connect/#/itc9267a2f12';
   audioVersionOptions: string[][];
+  hasStories: boolean;
 
   tabSub: Subscription;
   state: string;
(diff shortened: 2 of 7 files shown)


src/app/series/directives/series-podcast.component.html
```diff
@@ -49,7 +49,8 @@
       </div>
     </publish-fancy-field>
 
-    <publish-fancy-field label="Podcast Audio" [select]="audioVersionOptions" [model]="distribution" name="versionTemplateUrl" required>
+    <publish-fancy-field required label="Podcast Audio" name="versionTemplateUrl"
```
gcampo88: getting `Error in ./SeriesPodcastComponent class SeriesPodcastComponent - inline template:52:60 caused by: Cannot read property '0' of undefined` when I try to create a podcast in a new series. Looks like it's coming from this line in `get versionTemplateConfirm()` : `this.audioVersionOptions.find(opt => opt[1] === url)[0];` 
When I poke around in the debugger,`url` is an empty string at that point.

cavis: Well that's odd.  Will fix that!

cavis:
@gcampo88 give that a shot

gcampo88:
@cavis this means I don't get that error, but looks like the url from `let url = this.distribution.versionTemplateUrl;` is still an empty string right after I click "Save" on a newly created podcast. So the "podcast audio" select defaults to the only one I've created when I'm first creating my new podcast, but then once I've saved it, it goes to a blank option.

cavis:
You mean i was supposed to actually test this or something?

cavis:
@gcampo88 wasn't `encode()`-ing that param unless it looked changed.  And it didn't, since I force-set it.

It does seem right to not have the orange "changed" indicator on the field when we're defaulting it.  Downside is that it doesn't persist to localstorage ... so if you click the "Create Podcast" and then refresh the page, it's no longer selected by default.  Edge case!

gcampo88:
looks good! 👍  
I'll approve but won't merge yet in case @sandikbarr wants to take a look at your AdvancedConfirm changes

src/app/series/directives/series-podcast.component.ts
```diff
@@ -117,4 +122,17 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
...
   }
 
+  get versionTemplateConfirm(): string {
+    if (this.distribution && this.audioVersionOptions) {
+      let url = this.distribution.versionTemplateUrl;
+      let match = this.audioVersionOptions.find(opt => opt[1] === url);
+      let name = match ? match[0] : ''; // options are [[display, value]]
+      return `
+        Are you sure you want to use <b>${name}</b> as the audio for your podcast?
+        Changing this field can have catastrophic consequences for any previously
```
sandikbarr: Instead of "catastrophic consequences" explaining what it actually does would be more helpful. "This will change the audio version your previously published episodes are using."

sandikbarr: Also seems like scare tactics aren't necessary when there are 0 episodes in the series. This can be controlled outside the confirm directive by having an empty string for the confirmation message.

cavis: Good call.

You don't think we need to frighten our users into subservience?

sandikbarr: We are living in Trump's America after all...

cavis:
@sandikbarr updated!

Issue #196, opened 2017-01-10
Podcast, list of episodes table styles


kookster:
![image](https://cloud.githubusercontent.com/assets/46439/21811457/bcd5a6e4-d71d-11e6-8e8e-e59eaa9bfdca.png)

I think we should do a UI review of this tab.
Seems like it could look more like a table, or rows of data.

sandikbarr:
Kind of rough... I'm thinking accordions for each section with relevant links below the tabs
<img width="733" alt="screen shot 2017-01-12 at 8 27 08 pm" src="https://cloud.githubusercontent.com/assets/2250413/21916505/bbfe1ef0-d906-11e6-8b34-c4213b7381e6.png">

PR #309, opened 2017-01-27
episodes in series table presentation

sandikbarr:
#196


src/app/series/directives/series-feed.component.css
```diff
@@ -2,27 +2,39 @@ h4 {
...
+}
+
+li:last-child {
+  border-bottom: none;
 }
 
-.public {
-  color: #000000;
+li h5 {
+  width: calc(3 * 100%/5);
```
gcampo88: can get the same effects by using `li h5 { flex: 3 }` and `li p { flex: 1}` and skipping width, right?

Issue #291, opened 2017-01-24
series, audio templates confirm changes to templates with existing episodes


sandikbarr:
Changes to series audio templates could potentially affect the validity of existing episodes using those templates. If the user makes changes to an audio template that is in use in existing episodes, they should be shown the advanced fields confirmation modal with an explanation of how this could affect their episodes.

PR #316, opened 2017-01-30
asks user to confirm template changes

sandikbarr:
#291 Other than length, I'm not entirely sure how all these changes affect the validity of existing stories. For example, if I remove a segment from the template, stories still have that segment. I probably need some guidance here on appropriate instruction for the user for what we are asking them to confirm.

### DIFF
--- a/src/app/series/directives/series-templates.component.ts
+++ b/src/app/series/directives/series-templates.component.ts
@@ -55,7 +55,7 @@ import {
               </div>
               <publish-file-template *ngFor="let t of v.fileTemplates" [file]="t" [version]="v"></publish-file-template>
               <button tabindex=-1 class="add-segment" *ngIf="canAddFile(v)" type="button"
-                (click)="promptToAddFile(v)"><i class="icon-plus"></i>Add Segment</button>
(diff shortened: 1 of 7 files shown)


src/app/series/directives/series-templates.component.ts
```diff
@@ -51,8 +54,8 @@ import {
                 to validate the specific file.
               </div>
               <publish-file-template *ngFor="let t of v.fileTemplates" [file]="t" [version]="v"></publish-file-template>
-              <button tabindex=-1 class="add-segment" *ngIf="canAddFile(v)"
-                (click)="addFile(v)"><i class="icon-plus"></i>Add Segment</button>
+              <button tabindex=-1 class="add-segment" *ngIf="canAddFile(v)" type="button"
```
sandikbarr: This looks like the same bug as the one with the delete modal, so it may not get fixed in this PR.

sandikbarr: I did figure out a way to fix it for this PR after all

sandikbarr:
ok, back to you @gcampo88

PR #348, opened 2017-02-03
pop up a toast message when series or episode is created, updated, or deleted

sandikbarr:
#332

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "prosemirror-view": "^0.14.1",
     "pug": "^2.0.0-beta6",
     "rxjs": "5.0.0-beta.12",
-    "toastr": "^2.1.2",
     "ts-helpers": "^1.1.1",
     "url-template": "^2.0.8",
     "zone.js": "^0.6.23"
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "prosemirror-view": "^0.14.1",
     "pug": "^2.0.0-beta6",
     "rxjs": "5.0.0-beta.12",
+    "toastr": "^2.1.2",
     "ts-helpers": "^1.1.1",
     "url-template": "^2.0.8",
     "zone.js": "^0.6.23"
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -15,3 +15,4 @@
 <publish-footer></publish-footer>
 <publish-auth></publish-auth>
 <publish-modal></publish-modal>
+<publish-toastr></publish-toastr>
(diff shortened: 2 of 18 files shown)


gcampo88:
Code looks great! I might vote for having the toaster more centered on the page (I'd been unconsciously picturing it as taking up the full space of the `publish-header` when shown). And I think we could potentially shorten the message that's popped up, to skip the name of the series or episode -- thinking of long episode names:
screen shot 2017-02-03 at 5 09 31 pm

We could consider just having the text be `{Saved/ Deleted/ Created}!`

But I'm also happy to leave that kind of feedback to our testers!

gcampo88:
Also -- do we want to add this when an episode is published?

sandikbarr:
* 👍 yes to when published. even though we toggle the status, makes sense
* Last time I used generic toastr messages in an app user feedback said it was too ambiguous, which is why I included the title. Would centering the toast or some different formatting help or do long episode names just make the toasts too obtrusive?
* as far as positioning, I was thinking top right to be above the most often used hero buttons
* I feel like Episode/Series helps with ambiguity, esp when action that created the toast also navigates away like "Episode Deleted"
* should toasts have a visible close button in addition to esc to click?
* even though our design is very flat, do we want any sort of box shadow on this?

gcampo88:
- Would centering the toast or some different formatting help or do long episode names just make the toasts too obtrusive? -->Hard to say! Maybe we wait and see what user feedback is on it as is?
- I feel like Episode/Series helps with ambiguity, esp when action that created the toast also navigates away like "Episode Deleted" --> good point!
- should toasts have a visible close button in addition to esc to click? --> I don't see the specific need, since they're only there for 2.5 seconds and you can escape out of them... but that's just me. 
- even though our design is very flat, do we want any sort of box shadow on this? --> going to defer to @farski on this!

sandikbarr:
* went with "Episode deleted" etc (both words capitalized seemed off when I tried it)
* centered toasts
* added Episode published/unpublished toasts

sandikbarr:
@farski seems the open questions are positioning (center, right, somewhere else?) and whether or not to include a box shadow

gcampo88:
LGTM!

gcampo88:
@farski since you currently have no dev environment, here's what the toastr alerts currently look like  -- they fadein from the top of the screen and stay put for 2.5 sec, with different color for success/info/error/default. These are all from Series page but it looks same for Episode. Happens any time you Create/ Save/ Delete/ Publish. 

screen shot 2017-02-06 at 10 48 44 am


screen shot 2017-02-06 at 10 50 00 am


screen shot 2017-02-06 at 10 50 27 am


screen shot 2017-02-06 at 10 51 09 am

Issue #353, opened 2017-02-06
Episode, podcast distribution, set web link for the episode


kookster:
This would set the channel.item.link value in the rss feed, which is the feeder.episode.url

PR #357, opened 2017-02-07
Episode Web Link

gcampo88:
#353
(Also made tiny adjustment to Mock CMS Service to get rid of linting error in test logs.)

### DIFF
--- a/src/app/shared/model/feeder-episode.model.ts
+++ b/src/app/shared/model/feeder-episode.model.ts
@@ -74,21 +74,4 @@ export class FeederEpisodeModel extends BaseModel {
     }
   }
 
-  createLink(url: string): string {
-    let urlLength = url.length;
-    if (urlLength < 'https://'.length) {
-      if (['http://'.slice(0, urlLength), 'https://'.slice(0, urlLength)].indexOf(url) > -1) {
-        return url;
-      }
-    }
-    return /^https?:\/\//i.test(url) ? url : `http://${url}`;
-  }
-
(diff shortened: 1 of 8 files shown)


src/app/shared/model/feeder-episode.model.ts
```diff
@@ -69,4 +74,21 @@ export class FeederEpisodeModel extends BaseModel {
     }
   }
 
+  createLink(url: string): string {
```
sandikbarr: wonder if we should move that createLink stuff up to base model or out into a utility

src/app/story/directives/podcast.component.html
```diff
@@ -1,5 +1,10 @@
 <publish-spinner *ngIf="!episode"></publish-spinner>
 <form *ngIf="episode">
+  <publish-fancy-field label="Episode URL" textinput [model]="episode" name="webLink">
```
sandikbarr: I don't think this should be at the top. Seems like it will rarely be used. #338 doesn't exactly specify that explicit should be at the top, but it seems implied that it's the most often used field and should be at the top (above author info.) I don't imagine episode URL will be set frequently.

gcampo88: makes sense!

src/app/shared/model/feeder-episode.model.ts
gcampo88: good call!

gcampo88:
back to you @sandikbarr

PR #359, opened 2017-02-07
Require category & explicit for podcast

### LINKED ISSUE #318: Series new-podcast not validating right away
1. Navigate to the "Podcast Distribution" tab for a Series without one yet
2. Add a new podcast dist
3. Saved button is enabled even though required fields are blank (category, explicit, etc)
4. Refresh page without filling anything out or saving
5. Saved button is now disabled, as expected

gcampo88:
fixes #318

### DIFF
--- a/src/app/shared/model/feeder-podcast.model.ts
+++ b/src/app/shared/model/feeder-podcast.model.ts
@@ -24,6 +24,8 @@ export class FeederPodcastModel extends BaseModel {
   hasPublicFeed: boolean = false;
 
   VALIDATORS = {
+    category: [REQUIRED()],
+    explicit: [REQUIRED()],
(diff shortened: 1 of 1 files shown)


src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -24,6 +24,7 @@ export class FeederPodcastModel extends BaseModel {
   hasPublicFeed: boolean = false;
 
   VALIDATORS = {
+    category: [REQUIRED()],
     link: [REQUIRED(), URL('Not a valid URL')],
```
sandikbarr: let's add `explicit: [REQUIRED()],` here too, empty string/option should not be considered a valid value

gcampo88:
back to you @sandikbarr!

PR #362, opened 2017-02-08
allows user to select account under which to create series

### LINKED ISSUE #327: Create series in specific accounts when a user has access to multiple
If a user belongs to only a single account, there is no need to show or change anything in the UI, and we should continue to set the account for a series on that single account.

If there are multiple accounts, then list them by name in a select on the series, with the default account selected on a new series (as per prior conversations with @sandikbarr, no empty choice in the select).

This should be a required field.

For episodes of a series, it is not necessary to have them select which account they belong to - that is determined and updated based on the series: PRX/cms.prx.org#247

sandikbarr:
fixes #327


src/app/shared/model/series.model.ts
```diff
@@ -19,8 +19,9 @@ export class SeriesModel extends BaseModel implements HasUpload {
   public images: ImageModel[] = [];
   public versionTemplates: AudioVersionTemplateModel[] = [];
   public distributions: DistributionModel[] = [];
+  public accountId: number;
 
-  SETABLE = ['title', 'description', 'shortDescription', 'hasUploadMap'];
+  SETABLE = ['title', 'description', 'shortDescription', 'hasUploadMap', 'accountId'];
```
sandikbarr: accountId field is not persisted, just a place to manage it. I needed a model to stash it on to make all this work together though. Account doesn't quite make sense to me as a related field, but maybe that's how I should have done it.

src/app/series/series.component.spec.ts
```diff
@@ -30,13 +30,13 @@ describe('SeriesComponent', () => {
...
+    auth.mock('prx:series', {id: 99, title: 'my series title'}).mock('prx:account', {id: 78});
     fix.detectChanges();
     expect(el).toContainText('my series title');
     expect(comp.series.id).toEqual(99);
   });
 
-  cit('creates new series from the default account', (fix, el, comp) => {
+  cit('defaults new series to the default account', (fix, el, comp) => {
     activatedRoute.testParams = {};
     auth.mock('prx:default-account', {id: 88});
```
gcampo88: may want to add another account here to improve the test?

sandikbarr: Rewording this test is probably misleading. Honestly probably could have left the existing wording because it is still technically true. This isn't really where it uses multiple accounts. For New Series, series component will only have the default account and is used to instantiate the series, which is where it gets the default, but the multiple accounts come into play on the series basic component, which calls back to the series component to setSeries with the alternate account.

gcampo88: Ah, I gotcha.

src/app/series/directives/series-basic.component.spec.ts
```diff
@@ -1,22 +1,44 @@
...
   provide(TabService);
+  provide(Router, router);
+  provide(ActivatedRoute, activatedRoute);
+  provide(ModalService);
+  provide(ToastrService, {success: () => {}});
+  provide(SeriesComponent);
+
+  beforeEach(() => {
+    let auth = cms.mock('prx:authorization', {});
+    auth.mockItems('prx:accounts', [
```
gcampo88: might be worth it to add a test to check that it defaults here to default account, and/or a test to see that user is given option of changing accounts?

src/app/series/directives/series-basic.component.spec.ts
```diff
@@ -1,22 +1,44 @@
-import { cit, create, provide, By } from '../../../testing';
+import { cit, create, provide, By, cms } from '../../../testing';
+import { RouterStub, ActivatedRouteStub } from '../../../testing/stub.router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { SeriesBasicComponent } from './series-basic.component';
+import { SeriesComponent } from '../series.component';
+import { ModalService, ToastrService, CmsService } from '../../core';
```
gcampo88: getting a linter error -- unused import CmsService

src/app/series/directives/series-basic.component.spec.ts
sandikbarr: I believe the series.component test `defaults new series to the default account` covers the default to default account. I added a test to look for the account field on new series.

sandikbarr:
back to you @gcampo88

PR #364, opened 2017-02-10
Allow users to change accounts for existing series

### LINKED ISSUE #363: Update the account for a series
currently, a user with access to multiple accounts can set the account for a series on create.

add to this the ability to update and move the series to a different account to which they have access.

this is the kind of thing we should annoy/warn people about, as it will change who has access to the account (people who could access it before in the current account may very well be unable to edit it), and potentially who gets paid or pays for it (which is the other point of accounts - access and $).

Depends on completion of this ticket: PRX/cms.prx.org#252

gcampo88:
fixes #363

### DIFF
--- a/src/app/series/directives/series-basic.component.html
+++ b/src/app/series/directives/series-basic.component.html
@@ -1,0 +1,36 @@
+<form *ngIf="series">
+  <publish-fancy-field *ngIf="accountOptions?.length > 1" label="Owner" required>
+    <publish-fancy-field name="accountId" [select]="accountOptions" [model]="series"
+      [advancedConfirm]="newOwnerConfirm(series['accountId'])" small=1>
+      <div class="fancy-hint">
+        Who is the owner of your series?
+      </div>
+    </publish-fancy-field>
+  </publish-fancy-field>
+
+  <publish-fancy-field textinput required [model]="series" name="title" label="Series Title" required>
+    <div class="fancy-hint">What's the name of your series?</div>
+  </publish-fancy-field>
+
+  <publish-fancy-field textinput required [model]="series" name="shortDescription" label="Teaser" required>
+    <div class="fancy-hint">A short description of your series.</div>
+  </publish-fancy-field>
+
+  <publish-fancy-field label="Description">
+    <div class="fancy-hint">A full description of your series.</div>
+    <publish-wysiwyg [model]="series" name="description" [content]="series.description"
+      [images]="series.images" [changed]="descriptionChanged"></publish-wysiwyg>
+  </publish-fancy-field>
+
+  <publish-fancy-field label="Profile Image">
+    <div class="fancy-hint">This image will be used as the cover image for your series.</div>
+    <publish-image-upload [model]="series" purpose="profile" minWidth=1400 minHeight=1400></publish-image-upload>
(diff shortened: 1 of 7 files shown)


src/app/shared/model/series.model.ts
```diff
@@ -36,9 +37,8 @@ export class SeriesModel extends BaseModel implements HasUpload {
   constructor(account: HalDoc, series?: HalDoc, loadRelated = true) {
     super();
     this.init(account, series, loadRelated);
-    if (account) {
```
sandikbarr: I dunno why or if we need to support this, but the series.component used to init the SeriesModel with a null account. I changed that one and don't see it anywhere else, but if account is null/undefined and not checked, this will throw an error.

src/app/shared/model/series.model.ts
```diff
@@ -120,6 +120,9 @@ export class SeriesModel extends BaseModel implements HasUpload {
     data.title = this.title;
     data.descriptionMd = this.description;
     data.shortDescription = this.shortDescription;
+    if (this.changed('accountId')) {
+      data.set_account_uri = `api/v1/accounts/${this.accountId}`;
```
sandikbarr: should we be able to expand the existing account link and replace the id with the new account id? I'd feel better about doing that rather than hard coding anything for the API.

src/app/series/directives/series-basic.component.html
```diff
@@ -0,0 +1,36 @@
+<form *ngIf="series">
+  <publish-fancy-field *ngIf="accountOptions?.length > 1" label="Owner" required>
+    <publish-fancy-field name="accountId" [select]="accountOptions" [model]="series"
+      [advancedConfirm]="newOwnerConfirm(series['accountId'])" small=1>
+      <div class="fancy-hint">
+        Who is the owner of your series?
```
gcampo88: 👍

src/app/series/directives/series-basic.component.spec.ts
```diff
@@ -46,16 +49,19 @@ describe('SeriesBasicComponent', () => {
...
+    comp.series = {
+      changed: () => false,
+      account: {name: 'DefaultAccount', id: '222'}
+    };
     fix.detectChanges();
     expect(el).toQuery('publish-image-upload');
     expect(el).toContainText('cover image');
   });
 
   cit('when user has access to more than one account, allows setting owner for new series', (fix, el, comp) => {
```
gcampo88: good catch

src/app/series/directives/series-basic.component.ts
```diff
@@ -64,27 +27,28 @@ export class SeriesBasicComponent implements OnDestroy {
     });
   }
 
-  setAccount(accountId: string) {
```
gcampo88: 👍

src/app/shared/model/series.model.ts
gcampo88: belt + suspenders, right? :)

src/app/shared/model/series.model.ts
gcampo88: good call!

gcampo88:
back to you @sandikbarr

PR #365, opened 2017-02-13
adds google analytics for staging and production

### LINKED ISSUE #350: Google Analytics integration
as discussed with @cavis 
https://github.com/angulartics/angulartics2 supports ng2

sandikbarr:
fixes #350 

You will need to `rm -rf node_modules` and `npm install`


src/assets/google-analytics.js
```diff
@@ -1,6 +1,21 @@
...
-ga('create', 'UA-164824-57', 'auto');
+if (window.location.host.match(/prx.org/)) {
+  initGoogleAnalytics(window, document, 'script', 'https://www.google-analytics.com/analytics.js', 'ga');
+
+  ga('create', 'UA-164824-57', 'auto');
+} else if (window.location.host.match(/prx.tech/)) {
+  initGoogleAnalytics(window, document, 'script', 'https://www.google-analytics.com/analytics.js', 'ga');
+
+  ga('create', 'UA-164824-58', 'auto');
+}
```
sandikbarr: 'UA-' string correspond to Google Analytics ids for production and staging, none for dev

src/app/story/directives/status.component.ts
```diff
@@ -156,6 +158,7 @@ export class StoryStatusComponent implements DoCheck {
   togglePublish() {
     this.isPublishing = true;
     this.story.setPublished(!this.story.publishedAt).subscribe(() => {
+      this.angulartics2.eventTrack.next({ action: this.story.publishedAt ? 'publish' : 'unpublish', properties: { category: 'episode', label: 'episode/' + this.story.doc.id }});
```
gcampo88: linter's yelling at me about line length here

gcampo88: also, `this.story.publishedAt` is used to determine publish/unpublish 3x in 3 lines -- could save as a local var?

src/assets/google-analytics.js
gcampo88: 👍 
(I found this pretty helpful in understanding their code: http://code.stephenmorley.org/javascript/understanding-the-google-analytics-tracking-code/)

gcampo88: I see in their docs that they have this line anyway for localhost-- any reason for us to do that too? 
`ga('create', 'UA-XXXXXXXX-X', 'none'); // 'none' while you are working on localhost`
Does it help with testing? Hunting around in docs but I assume you're quite familiar with them at this point :)

sandikbarr: Looking for documentation on what 'none' as the cookieDomain actually does...  https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#cookieDomain I assume 'none' is for when you are using localhost rather than faking a domain like we do with pow. Guess I could just try it. Thought we didn't really want this locally at all though. I don't want to take up another of our 11 Google Analytics properties we have left for our local dev environments, so just re-use the staging id as long as it's not actually posting page views or events.

src/app/story/directives/status.component.ts
```diff
@@ -156,6 +158,8 @@ export class StoryStatusComponent implements DoCheck {
   togglePublish() {
     this.isPublishing = true;
     this.story.setPublished(!this.story.publishedAt).subscribe(() => {
+      this.angulartics2.eventTrack.next({ action: this.story.publishedAt ? 'publish' : 'unpublish',
+        properties: { category: 'episode', label: 'episode/' + this.story.doc.id }});
```
sandikbarr: can't save in a local variable and use in all three places cuz it changes with the publish/unpublish. gonna keep using the date as boolean because it seems like there are some pretty strong feelings about that.

src/assets/google-analytics.js
sandikbarr: Calling ga 'create' with 'none' as the cookieDomain tracks local traffic. We don't want those mixed up with staging or production, so sticking with nothing for local.

sandikbarr:
back to you @gcampo88

PR #369, opened 2017-02-28
Upgrade angular cli

cavis:
Upgrade the angular cli.

- [x] Upgrade to angular-cli rc0
- [x] Simplify docker build, since now it can use a prebuilt sass binary
- [x] Linting

Hacks to make this work:

- Locked the router version, as the next version breaks us #370
- Couple of non-default changes to the polyfills and tsconfig.spec, to get phantom working


tslint.json
```diff
@@ -61,6 +63,7 @@
       "check-else",
       "check-whitespace"
     ],
+    "prefer-const": false,
```
cavis: Should set this back to `true` at some point.  But probably in its own PR, as it flags a lot of stuff.

tslint.json
```diff
@@ -11,12 +12,14 @@
     "curly": true,
     "eofline": true,
     "forin": true,
+    "import-blacklist": [false, "rxjs"],
```
cavis: This should also be set back to true.  But it requires changing every single `from 'rxjs'` import.  And figuring out exactly what operators we need.  Apparently does trim down the size of that lib quite a bit, though.

src/tsconfig.spec.json
```diff
@@ -0,0 +1,26 @@
...
+    "declaration": false,
+    "moduleResolution": "node",
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "lib": [
+      "es2016"
+    ],
+    "outDir": "../out-tsc/spec",
+    "module": "commonjs",
+    "target": "es5",
```
cavis: Originally this was `es6` - but had to make some changes to get Phantom to work.  Not sure what impact this will have ... maybe bad stack traces when running the specs?  Not sure.

src/polyfills.ts
```diff
@@ -1,5 +1,24 @@
...
+ *
+ * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
+ */
+
+/***************************************************************************************************
+ * BROWSER POLYFILLS
+ */
+
+/** IE9, IE10 and IE11 requires all of the following polyfills. **/
 import 'core-js/es6/symbol';
```
cavis: These core-js/es6 polyfills were all commented out, when generating a new cli project.

Had to uncomment them to get phantom working.  Not entirely sure of the impact here ... slightly larger dist js?

tslint.json
sandikbarr: That is a lot. Noted.

PR #374, opened 2017-03-06
make episode url (feeder item link) required

### LINKED ISSUE #371: Story Episode URL is required
This field seems to be defaulting to the story.alternate link.  Which is great!

Shouldn't let users unset this field, though.  It causes the feed `<item><link/>` to be blank.  Which isn't valid.

sandikbarr:
fixes #371


src/app/story/story.component.ts
```diff
@@ -98,6 +98,9 @@ export class StoryComponent implements OnInit {
     } else if (this.story) {
       this.story.loadRelated('distributions').subscribe(() => {
         this.distPodcast = this.story.distributions.some(d => d.kind === 'episode');
+        if (this.distPodcast) {
+          this.story.distributions.find(d => d.kind === 'episode').loadRelated('episode');
```
sandikbarr: I found that if I did not loadRelated on the episode up front, the status box would not show that the episode URL field is required until navigating to the Podcast Distribution tab.

cavis: Good catch!

PR #383, opened 2017-03-14
Fix wysiwyg discard

cavis:
Tries creating a new state on every change.  (Except value changes originating from the editor itself... those do not trigger anything).

Should fix some issues around discarding/saving and adding/removing images.

### DIFF
--- a/src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
+++ b/src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
@@ -20,29 +20,21 @@ export class ProseMirrorImage {
 export class ProseMirrorMarkdownEditor {
 
   view: MenuBarEditorView;
-  savedState: EditorState;
 
   constructor(private el: ElementRef,
-              private value: string,
+              public value: string,
               private images: ProseMirrorImage[],
               private setModel: Function,
               private promptForLink: Function) {
-    this.savedState = EditorState.create(this.stateConfig());
-    this.view = new MenuBarEditorView(el.nativeElement, this.viewProps(this.savedState));
+    let state = EditorState.create(this.stateConfig());
+    this.view = new MenuBarEditorView(el.nativeElement, this.viewProps(state));
   }
 
-  update(images: ProseMirrorImage[]) {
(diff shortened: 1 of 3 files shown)


src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -58,7 +52,8 @@ export class ProseMirrorMarkdownEditor {
       state,
       onAction: (action) => {
         this.view.updateState(this.view.editor.state.applyAction(action));
-        this.setModel(defaultMarkdownSerializer.serialize(this.view.editor.state.doc));
+        this.value = defaultMarkdownSerializer.serialize(this.view.editor.state.doc);
```
kookster: I see holding onto the state for comparison in `update`

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -20,29 +20,23 @@ export class ProseMirrorImage {
...
-    this.view.update(this.viewProps(state));
-  }
-
-  resetEditor() {
-    this.view.updateState(this.savedState);
-  }
-
-  setSavedState() {
-    this.savedState = this.view.editor.state;
+  update(value: string, images?: ProseMirrorImage[]) {
```
kookster: signature change, getting images array too.
not comparing the existing images to this, always just overwriting when passed in?
probably all that does is rebuild the menu, will have to check...

kookster: yup, in  stateConfig calls buildMenuItemsPlugin, and that iterates through the images

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -20,29 +20,23 @@ export class ProseMirrorImage {
...
+    if (value !== this.value || images) {
+      this.value = value;
+      this.images = images || this.images;
```
kookster: I see, the ngOnChanges logic only passes in the `images` when they have been changed, so no need to compare them.

PR #385, opened 2017-03-14
Error reporting

cavis:
- [x] Get NR browser headers set correctly (again)
- [x] Add runtime ENVs for new relic and google analytics
- [x] Demo error handling page


cavis:
I wasn't getting an error modal popping up on runtime errors.  I'm not sure why 3ddea67 doesn't work for this case, but I reverted that, and added manual change detection back in.

As far as i can tell, that didn't break anything.  (No double-modals seen so far).

Issue #317, opened 2017-01-31
Additional attributes that can be set for a podcast


kookster:
These are fields we support in feeder for the RSS of the channel, but I don't think we have exposed in the publish UI. Most of these are not that important, but consider adding them to the bottom of the podcast attributes, and for some (like complete) to the advanced section.

- [ ] managing editor - optional, don't default since not required
- [ ] web_master - similar to above
- [ ] copyright - optional, can probably default this in some way
- [ ] complete (should be advanced) - indicates no more episodes will be published
- [ ] language - currently defaults to 'en-us', provide way to set this, advanced?
- [ ] `sy` tags ttl, update_frequency, update_period - utterly optional, not really used by apps AFAIK, advanced?

kookster:
webmaster not in the feeder API, so need to see about that before adding here.

kookster:
these fields aren't terribly important - icebox

PR #386, opened 2017-03-15
Add podcast fields

### LINKED ISSUE #343: series, podcast distribution unable to clear author email
The series podcast distribution author email field cannot be cleared after it has been set. As this is not a required field, empty string should be allowed.

kookster:
Working on #317 

To the podcast info form:
- [x] Add `copyright`
- [x] Add `complete` 
- [x] Advanced confirm on `complete`
- [x] Add `language`
- [x] Add `managingEditor`
- [x] Add `owner`

To make this happen:
- [x] Add checkbox to fancy field
- [x] Get a list of languages to select from
- [x] Write a better advanced confirm message for `complete`

also fixes #343

### DIFF
--- a/src/app/series/directives/series-podcast.component.ts
+++ b/src/app/series/directives/series-podcast.component.ts
@@ -122,6 +122,13 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
     }
   }
 
+  get completeConfirm(): string {
+    if (this.podcast) {
+      let confirmMsg = 'Are you sure you want to set this as complete, and there will be no more episodes in the podcast?';
+      return confirmMsg;
+    }
+  }
+
   get versionTemplateConfirm(): string {
     if (this.distribution && this.audioVersionOptions && this.series.hasStories) {
       let url = this.distribution.versionTemplateUrl;
--- a/src/app/series/directives/series-podcast.component.ts
+++ b/src/app/series/directives/series-podcast.component.ts
@@ -2,6 +2,7 @@ import { Component, OnDestroy, DoCheck } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { SeriesModel, DistributionModel, FeederPodcastModel,
          TabService, CATEGORIES, SUBCATEGORIES } from '../../shared';
+import * as languageMappingList from 'langmap';
 
 @Component({
   styleUrls: ['series-podcast.component.css'],
@@ -24,8 +25,10 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
   initialDistributions: DistributionModel[];
   distribution: DistributionModel;
   podcast: FeederPodcastModel;
+  languageOptions: string[][];
 
   constructor(tab: TabService) {
+    this.languageOptions = this.getLanguageOptions();
     this.tabSub = tab.model.subscribe((s: SeriesModel) => {
       this.series = s;
       this.series.loadRelated('versionTemplates').subscribe(() => {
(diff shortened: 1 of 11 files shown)


kookster:
@sandikbarr I finally finished a publish PR! pls review

src/app/series/directives/series-podcast.component.ts
```diff
@@ -122,6 +137,13 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
     }
   }
 
+  get completeConfirm(): string {
+    if (this.podcast && this.podcast.complete) {
+      let confirmMsg = 'Are you sure this podcast is complete? Apps will assume this is show over, and won\'t look new episodes.';
```
sandikbarr: is this missing a word? should this say "won't look _for_ new episodes"?

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -87,28 +87,44 @@ export class FeederPodcastModel extends BaseModel {
...
-    if (this.authorName || this.authorEmail) {
-      data.author = {
-        name: this.authorName,
-        email: this.authorEmail
-       };
-    }
-
-    data.enclosurePrefix = this.enclosurePrefix || null;
+    ['author', 'owner', 'managingEditor'].forEach((role) => {
+      if (this[`${role}Name`] || this[`${role}Email`]) {
```
sandikbarr: If these fields are empty string, the old values are retained. You can clear one of the fields, name or email, but not both. There's an existing bug ticket on this for authorEmail #343. We can either fix here or update that ticket with these additional fields.

kookster: The reason we made author email required was b/c we were using it for the owner values (I believe), as itunes need an owner to be able to claim a feed, and author is optional in rss.

Now that we have owner, we should probably change the validation, and shift it to owner - I'll take a look.

kookster: ok, misread he bug, I see the problem, and I'll put in a fix so these values can be blanked out.

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -87,28 +87,42 @@ export class FeederPodcastModel extends BaseModel {
...
-    if (this.authorName || this.authorEmail) {
-      data.author = {
-        name: this.authorName,
-        email: this.authorEmail
-       };
-    }
-
-    data.enclosurePrefix = this.enclosurePrefix || null;
+    ['author', 'owner', 'managingEditor'].forEach((role) => {
+      data[role] = {
```
kookster: this should fix #343

kookster:
I think I fixed those issues @sandikbarr - good call on #343

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -120,6 +134,12 @@ export class FeederPodcastModel extends BaseModel {
     } else {
       data.itunesCategories = [];
     }
+
+    data.explicit = this.explicit || null;
```
sandikbarr: same comment on name/email, but since the debate we had was specifically about explicit commenting here- don't 🔫 the messenger, but @cavis has told me in the past that we set empty values to empty string and not to use `null`

kookster: I only moved this line of code to alphabetize the attributes being set (yes, I am crazy).
In `encode`, it seems to be these get set to `null`, and in `decode` to empty string `''`, which makes sense to me as we don't want to save empty string values in the API when it should be a `null`.
@cavis can you confirm if `null` or empty string `''` is appropriate here?

PR #387, opened 2017-03-15
Get phantom working in alpine

### LINKED ISSUE #26: docker development mode
Current docker setup only supports serving precompiled assets.
- Make sure docker-compose correctly mounts dev `node_modules` back into the container.
- If it turns out we need `node-sass` in the container, may need to do something drastic.  Like caching the built binary somewhere, and moving it back when it's needed.
- Install `watchman`
- Map reload ports so webpack dev server works properly.

cavis:
Fixes #26.

- [x] Get phantomjs working in alpine
- [x] Get docker-compose tests working
- [x] Get docker-compose dev server working

Finally found an alpine+phantom project that works.  And doesn't require building from source.  Not quite as slick as a selenium/webdriver solution, but after struggling with that for awhile, I gave up.

This will install dev/optional dependencies into the docker container.  Which results in a slightly larger image.  But much better than non-alpine builds, which were pushing 1GB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,9 @@
 .gitignore
 coverage
 dist
+Dockerfile
 docker-compose.yml
+README.md
 node_modules
 tmp
 typings
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,19 +6,22 @@ ENV TINI_VERSION v0.9.0
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static /tini
 RUN chmod +x /tini
 
-WORKDIR /app
+ENV APP_HOME /app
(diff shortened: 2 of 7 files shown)


kookster:
looks like id doesn't love .docker, did you have it running / browsing locally?

image

cavis:
@kookster did you change your AUTH_CLIENT_ID?  (I think I noted that in the README ... it defaults to a prx.dev one right now)

cavis:
Oh wait ... hold on this one.  Dev reloading not actually working.  (It's trying to hit `localhost:4201` from the browser ... rather than proxying that through `publish.prx.docker`).

cavis:
Back in business!

PR #391, opened 2017-03-28
Better buildspec

cavis:
Push the image built by docker-compose to ECR.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,23 +1,26 @@
-publish:
-  build: .
-  volumes:
-    - ./bin:/app/bin
-    - ./e2e:/app/e2e
-    - ./lib:/app/lib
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,23 +1,26 @@
...
-    - "4200:4200"
-    - "4201:4201"
-  command: start
-  environment:
-    VIRTUAL_HOST: publish.prx.docker
+version: "2"
+services:
+  publish:
+    build: .
+    image: publish_app
```
cavis: This is why i had to upgrade to the v2 docker-compose.yml format.  The `image:` lets you explicitly tell it what docker tag to use.  Otherwise, it would get named `{working_dir_name}_publish`.

PR #392, opened 2017-04-05
podcast and episode iTunes summary

### LINKED ISSUE #378: Set the episode summary
Currently we use the html description in both the <description> tag and the <itunes:summary> tag, with the summary stripped of all but anchor tags.

Some folks still like to edit this themselves, so we could offer that as an option on the `podcast episode info` tab.

Could that be a stripped down prosemirror editor with only links available as formatting?

sandikbarr:
fixes #378 

iTunes summary is a new kind of wysiwyg editor that only allows plain text and links, no line breaks are allowed in this plain text. When summary has not been set, it is populated from the description field with all formatting but links removed. If the user changes the text in the summary wysiwyg editor, the new summary will be saved to the iTunes summary field; otherwise, summary is not set as it will be populated in the feed from the description.


src/app/series/directives/series-podcast.component.html
```diff
@@ -85,6 +85,20 @@
       </div>
     </publish-fancy-field>
 
+    <publish-fancy-field label="Summary" *ngIf="podcast && podcast.summary">
+      <div class="fancy-hint">The iTunes summary for this podcast. You may create links, but other formatting is not supported.</div>
+      <publish-wysiwyg [model]="podcast" name="summary" [content]="podcast.summary"
+                       inputFormat="HTML" outputFormat="HTML" [changed]="summaryChanged"></publish-wysiwyg>
```
sandikbarr: note to self: probably don't need this getter

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -161,11 +225,8 @@ export class ProseMirrorMarkdownEditor {
   markActive(state, type) {
     let {from, to, empty} = state.selection;
     if (empty) {
-      let activeMark = type.isInSet(state.storedMarks || state.doc.marksAt(from));
-      // trick prosemirror into checking from + 1 in case the cursor is at the beginning of a mark
```
sandikbarr: prosemirror fixed this issue

src/app/series/directives/series-podcast.component.css
```diff
@@ -25,3 +25,6 @@
 .feed-url button:hover, .feed-url button:focus {
   background-color: #e28b42;
 }
+input[type=checkbox] {
+  vertical-align: middle;
+}
```
sandikbarr: noting that `vertical-align: middle` should probably be our default on checkboxes since we use it wherever we use checkboxes, one for the styleguide!

src/app/shared/wysiwyg/wysiwyg.component.css
```diff
@@ -10,64 +10,78 @@ div >>> .ProseMirror-menubar-wrapper {
...
+}
+
+div.readonly >>> .ProseMirror {
+  color: #939393;
+  background-color: #f0f0f0;
+  min-height: 0px;
+}
+
+div.readonly >>> .ProseMirror-menubar {
+  display: none;
```
sandikbarr: There is a separate view type, EditorView as opposed to the MenuBarEditorView, that doesn't provide a menu, but API is slightly different, so I chose to hide the menu with CSS instead.

src/app/story/directives/podcast.component.ts
```diff
@@ -92,4 +96,7 @@ export class PodcastComponent implements OnDestroy {
     }
   }
 
+  toggleAlternateSummary() {
```
gcampo88: getting `Cannot read property 'getContent' of undefined` for a podcast episode without a description set when I click the toggle checkbox 
image

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -88,49 +88,37 @@ export class ProseMirrorMarkdownEditor {
...
+  plainTextWithLinks() {
+    let content = '';
     const getContent = (node) => {
       if (node.type.name === 'text') {
-        if (markdownSchema.marks.link.isInSet(node.marks)) {
-          content.push(node);
+        if (content.length > 0 && node.textContent.length > 0
+          && !node.textContent.match(/^\s+/)
+          && !content.match(/^.+\s$/)) {
+          content += ' ';
```
gcampo88: regex rookie confirms: this is just inserting a space in case two nodes don't have one between each other, right?

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -17,72 +18,159 @@ export class ProseMirrorImage {
...
+  }
+
+  plainTextWithLinks() {
+    let content = '';
+    const getContent = (node: Node) => {
+      if (node.type.name === 'text') {
+        if (content.length > 0 && node.textContent.length > 0
+          && !node.textContent.match(/^\s+/)
+          && !content.match(/^.+\s$/)) {
+          content += ' ';
```
gcampo88: regex rookie confirms: this is just inserting a space in case two nodes don't have one between each other, right?
also, i'd vote something like `translatedContent` instead of `content` just to be super clear that that's the bit that's already clean.

src/app/story/directives/podcast.component.ts
sandikbarr: Good test, caught by the empty string falseyness. Will get a fix out asap.

sandikbarr: Fixed

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
```diff
@@ -246,153 +331,166 @@ export class ProseMirrorMarkdownEditor {
...
-      result.push(headingRule(schema.nodes.heading, 6));
+    if (this.outputSchema.nodes.heading) {
+      result.push(headingRule(this.outputSchema.nodes.heading, 6));
     }
     return result;
   }
 
   buildMenuItemsPlugin() {
-    let r = {};
+    let r = {}, deps;
```
gcampo88: maybe slightly more descriptive variable names (is r for render and deps for dependencies?) ?

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
sandikbarr: Yep, that's exactly what it is doing. When nodes are concatenated, there generally isn't space between them already, so this is adding those spaces in between paragraphs, list items, etc.

Good call on the rename. Done!

src/app/series/directives/series-podcast.component.css
gcampo88: +1 for styleguide. think we need this on the episode podcast.component.css too! 
image

src/app/shared/wysiwyg/prosemirror.markdown.editor.ts
sandikbarr: I don't know what r is TBH, but yeah I'm pretty sure that deps is dependencies. I took this code from a prosemirror example to set up the menu, so I'm gonna leave it as close to that as possible. Most of his code looks this way, sigh.

gcampo88: 👍

src/app/shared/wysiwyg/wysiwyg.component.css
gcampo88: 👍

src/app/series/directives/series-podcast.component.css
sandikbarr: doh. fixed. Not looking forward to cleaning up all these one offs. At least we don't have to do it all at once. I'm seeing some "the styleguide provides..." PR comments in our future. :)

src/app/shared/wysiwyg/prosemirror.markdown.editor.spec.ts
```diff
@@ -53,4 +60,22 @@ describe('ProseMirrorMarkdownEditor', () => {
       expect(pmEditor.view.update).toHaveBeenCalled();
     });
   });
+
+  describe('getContent', () => {
+    let pmEditor;
+    let content = 'hola, [buenos](https://en.wikipedia.org/wiki/Isla_Mujeres "Isla Mujeres") dias';
```
gcampo88: :taco: town!

src/app/series/directives/series-podcast.component.css
gcampo88: good call!

sandikbarr:
That clearing out the box and reverting back to the original description is intentional, but it is good to have your feedback on that behavior. I'm assuming the first thing you did was blank it out then think you were gonna type something totally brand new. I think we assumed they might make a few small changes but keep it mostly the same. Andrew and I originally discussed that empty description is not allowed, but adding a conditional REQUIRED validator to summary got pretty complicated. We decided to go with this behavior rather than figuring out the validation to show a message and have a separate toggle/action to revert back. Totally open to it, but I feel like we want more user feedback on this before trying something different.

Issue #390, opened 2017-03-28
story downloads chart


sandikbarr:
- [x] For a story, display a downloads chart for the episode. Shows up to the minute data.
- [x] User is able to select datetime ranges for charted data

cavis:
Done!

PR #395, opened 2017-04-18
HOLD episode downloads chart

sandikbarr:
HOLD #390 showing the episode downloads chart


src/app/core/cms/castle.service.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.replayRoot.next(obj);
+      },
+      (err) => {
+        this.replayRoot.error(err);
+      }
+    );
+  }
+
+  protected getRemote(): HalRemote {
+    return new HalRemote(this.http, Env.CASTLE_HOST);
```
cavis: So this is assuming no basic-auth required, right?

src/app/story/directives/metrics-downloads.component.ts
```diff
@@ -0,0 +1,164 @@
...
+          this.chartData = [new TimeseriesChartModel(dataset, this.story.title, '#61A85D')];
+        } else {
+          this.error = 'This podcast has no download metrics.';
+        }
+      });
+    }
+  }
+
+  checkRequest() {
+    if (!this.story || !this.story.publishedAt) {
```
cavis: I'm still getting an endless spinner, for an unpublished/draft episode.  No console errors, though.  And not making any request to castle.

sandikbarr: ah yes. will fix. the spinner also doesn't go away when you do make a request to castle and it gets a 500, seems like on no data

src/app/core/cms/castle.service.ts
sandikbarr: right. I had a basic auth header stuck into the proxy at one point, but I removed that proxy.

src/app/story/directives/metrics-downloads.component.ts
cavis: Oh, okay, so it's actually that my series has no podcast-distribution.

I'm just going to do a quick fix to hide the tab for new-or-podcast-less stories.  And throw up an error message if Castle returns a non-200.

gcampo88:
@cavis your changes look good! Only caveats: I'm getting `The <template> element is deprecated. Use <ng-template> instead`, and currently getting a 406 from castle on the download tab --  `(Phoenix.NotAcceptableError) no supported media type in accept header, expected one of ["json", "hal"]`. Have added a castle ticket (or do we want to just alter that in castle.service?) https://github.com/PRX/castle.prx.org/issues/9. 

Anywho, will +1 this as well but refrain from merging.

cavis:
Away we go!

PR #399, opened 2017-05-09
Castle basic auth

cavis:
Requires #395.

- [x] Proxy castle requests through localhost, and add basic-auth header
- [x] Add a `CASTLE_TEST_EPISODE` param which, when present, uses a fixed episode-guid for the castle requests instead of the actual one.  Lets us look at some actual metrics in staging.


cavis:
Pretty simple non-frontend PR.  Going to merge and see if i can't get this running in staging.

PR #402, opened 2017-05-17
Fallback to native validation

cavis:
If Aurora validation is taking too long (it fails on mp3s with large id3 images) ... fallback to native <audio> to get the type/duration.

Not always accurate, and lacks bitrate/frequency, but should work for podcast mp3s.


cavis:
Confirmed that tests still pass against this.

cavis:
Also confirmed that build/prod-server works as expected.

Issue #407, opened 2017-05-25
Unavailable Series Loads Forever


debenedictis:
When visiting a series that is unavailable or non-existent the page displays the _still loading_ logo indefinitely, e.g., https://publish.prx.org/series/666 .

It would be nice if instead there was some message to indicate that that the requested page is unavailable.

cavis:
Fix deployed to prod.

PR #412, opened 2017-06-05
handle 404s for series and story 

gcampo88:
#407

### DIFF
--- a/src/app/core/cms/index.ts
+++ b/src/app/core/cms/index.ts
@@ -4,4 +4,4 @@ export { HalDoc } from './haldoc';
 export { HalLink } from './hallink';
 export { HalObservable } from './halobservable';
 export { HalRemoteCache } from './halremote.cache';
-export { HalRemote } from './halremote';
+export { HalRemote, HalHttpError } from './halremote';
--- a/src/app/series/series.component.ts
+++ b/src/app/series/series.component.ts
@@ -47,7 +47,7 @@ export class SeriesComponent implements OnInit {
           if (err.status === 404 && err.name === 'HalHttpError') {
             this.toastr.error('No series found. Redirecting to new series page');
             console.error(`Series with id ${this.id} not found`);
-            window.setTimeout(() => this.router.navigate(['/series', 'new']), 3000);
+            setTimeout(() => this.router.navigate(['/series', 'new']), 3000);
           }
         }
       );
--- a/src/app/series/series.component.ts
+++ b/src/app/series/series.component.ts
@@ -48,6 +48,7 @@ export class SeriesComponent implements OnInit {
             this.toastr.error('No series found. Redirecting to new series page');
             console.error(`Series with id ${this.id} not found`);
             setTimeout(() => this.router.navigate(['/series', 'new']), 3000);
+            throw(err);
           }
         }
       );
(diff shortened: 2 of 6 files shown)


src/app/series/series.component.ts
```diff
@@ -37,11 +37,19 @@ export class SeriesComponent implements OnInit {
...
-        });
-      });
+      this.cms.auth.follow('prx:series', {id: this.id}).subscribe(
+        s => {
+          s.follow('prx:account').subscribe(a => {
+            this.setSeries(a, s);
+          });
+        },
+        err => {
+          if (err.status === 404 && err.name === 'HalHttpError') {
```
cavis: Does typescript not freak out about "unknown property: status" here?

gcampo88: nope! I wonder why...

src/app/core/cms/index.ts
```diff
@@ -4,4 +4,4 @@ export { HalDoc } from './haldoc';
 export { HalLink } from './hallink';
 export { HalObservable } from './halobservable';
 export { HalRemoteCache } from './halremote.cache';
-export { HalRemote } from './halremote';
+export { HalRemote, HalHttpError } from './halremote';
```
cavis: Note to self: PR #413 is going to have fun merging this.

src/app/series/series.component.ts
```diff
@@ -37,11 +37,20 @@ export class SeriesComponent implements OnInit {
...
```
cavis: I'm not sure checking `err.name` is the usual typescript way of doing this - does `err instanceof HalHttpError` work?  (Not a huge deal, though).

src/app/series/series.component.ts
```diff
@@ -37,11 +37,20 @@ export class SeriesComponent implements OnInit {
...
+            this.toastr.error('No series found. Redirecting to new series page');
+            console.error(`Series with id ${this.id} not found`);
+            setTimeout(() => this.router.navigate(['/series', 'new']), 3000);
+          }
```
cavis: If you didn't handle the error here, you should probably re-throw it so it doesn't just disappear.

src/app/story/story.component.ts
```diff
@@ -60,13 +60,21 @@ export class StoryComponent implements OnInit {
...
+            story.follow('prx:series').subscribe(series => this.setStory(series, story));
+          } else {
+            this.setStory(null, story);
+          }
+        },
+        err => {
+          if (err.status === 404 && err.name === 'HalHttpError') {
+            this.toastr.error('No episode found. Redirecting to new episode page');
+            setTimeout(() => this.router.navigate(['/story', 'new']), 3000);
+          }
```
cavis: Again, re-throw unhandled errors.

src/app/core/cms/index.ts
gcampo88: :)

src/app/series/series.component.ts
gcampo88: `err instanceof HalHttpError` does not work -- not sure why! But looks like we've run into that before --
 https://github.com/PRX/publish.prx.org/blob/master/src/app/story/directives/metrics-downloads.component.ts#L85

src/app/series/series.component.ts
gcampo88: 👍

src/app/story/story.component.ts
gcampo88: 👍

src/app/series/series.component.ts
cavis: May be because extending `Error` in typescript is a bit wonky in the first place.  Ah well... this works!

PR #413, opened 2017-06-05
Use styleguide HAL

cavis:
Use styleguide HAL services.  Requires PRX/styleguide.prx.org#21.

- [x] Use HAL service from styleguide (actually uses `HalBaseService` ... which in turn uses `HalService`)
  - Link between Auth and HAL is now implicit in the service, rather than the top-level AppComponent handling the token handoff.
- [x] Use HAL mocks from styleguide
- [x] Switch to Castle PRX-auth (requires PRX/castle.prx.org#16 to work ... but shouldn't block merging this PR)


src/app/core/hal/feeder.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { HalBaseService } from 'ngx-prx-styleguide';
+import { Env } from '../core.env';
+
+@Injectable()
+export class FeederService extends HalBaseService {
+
+  get host(): string {
+    return 'feeder.prx.org'; // TODO
```
cavis: We don't actually use FeederService yet, so left this as a TODO.  (Currently to get to feeder, we just navigate via an absolute-link found in a CMS series/story).

src/app/story/directives/metrics-downloads.component.spec.ts
```diff
@@ -8,41 +8,30 @@ describe('MetricsDownloadsComponent', () => {
...
-  });
+  const mockDownloads = [{
+    downloads: [
+      ['2017-04-19T00:00:00Z', 99],
+      ['2017-04-20T00:00:00Z', 97]
+    ]
+  }];
 
   cit('should not make castle request if episode not published', (fix, el, comp) => {
-    spyOn(castle, 'followList').and.stub();
```
cavis: This method no longer exists on the global testing `cms` var (which is renamed to `castle` here).  You can't directly access the root api haldoc.

But ... by not mocking anything here, we can be sure we'd get an "unmocked request" error if the component did try to follow a link.

src/testing/components.ts
```diff
@@ -36,18 +36,23 @@ export function cit(desc: string, testFn: ComponentTestCallback) {
...
 // create a component
 export function create(componentType: Type<any>, runInitialDetect = true) {
-  mockCms = new MockCmsService();
+  mockHal = new MockHalService();
   beforeEach(function() {
     this._publishComponent    = componentType;
     this._publishDeclarations = [componentType, StubRouterLinkDirective];
-    this._publishProviders    = [{provide: CmsService, useValue: mockCms}, {provide: CastleService, useValue: mockCms}];
+    this._publishProviders    = [
+      {provide: HalService, useValue: mockHal},
```
cavis: We're just mocking out the HalService with the mock one.  This way, we can just use the actual Castle/Cms/Feeder services, and dependency injection will provide the MockHalService to them.

gcampo88:
oops, hang on. tests failing after master merged in

PR #416, opened 2017-06-15
Add margin to p tags in wysiwyg editor

gcampo88:


src/app/shared/wysiwyg/wysiwyg.component.css
```diff
@@ -87,6 +87,15 @@ div >>> .ProseMirror h6 {
   margin-bottom: 2.33em;
 }
 
+div >>> .ProseMirror p {
+  margin-top: 1em;
+  margin-bottom: 1em;
+}
```
sandikbarr: perfect! Exactly what a browser should have as defaults in the user agent style sheet. https://www.w3schools.com/tags/tag_p.asp (Our reset css is zeroing out margins.)

PR #417, opened 2017-06-16
utilize global form CSS from styleguide

sandikbarr:
Goes with https://github.com/PRX/styleguide.prx.org/issues/27


src/app/search/directives/search-form.component.css
```diff
@@ -0,0 +1,33 @@
...
+  border: 0;
+  padding: 6px;
+  color: #939393;
+  background: #fff;
+}
+
+option[disabled] {
+  display: none;
+}
+
```
sandikbarr: When moving this .form-group stuff into here, I noticed these two css files were identical and consolidated.

src/app/series/directives/series-podcast.component.css
```diff
@@ -24,6 +24,3 @@
 .feed-url button:hover, .feed-url button:focus {
   background-color: #e28b42;
 }
-input[type=checkbox] {
```
sandikbarr: no longer have to override this reset

src/app/shared/form/fancy-field.component.css
```diff
@@ -53,34 +53,16 @@ p.error {
 input, textarea {
   width: 100%;
   padding: 10px 8px;
-  color: #939393;
```
sandikbarr: Stuff that was removed here was moved to the global form CSS scope

src/app/shared/image/image-upload.component.css
```diff
@@ -24,7 +24,6 @@
 }
 .new-image label {
   padding: 6px 10px;
-  margin-bottom: 0;
```
sandikbarr: .button now has `margin: 0` so doesn't have to clear label's `margin-bottom: 3px`

Issue #406, opened 2017-05-24
Yarn-ify


cavis:
Tired of dependencies upgrading themselves and breaking?  https://yarnpkg.com/en/

cavis:
Dones!

PR #418, opened 2017-06-20
Init yarn

cavis:
#406.  Yarn is more better.


Dockerfile
```diff
@@ -26,11 +26,14 @@ ENTRYPOINT [ "/tini", "--", "./bin/application" ]
 CMD [ "serve" ]
 
 ADD ./package.json ./
+ADD ./yarn.lock ./
 RUN apk --update add curl git && \
   curl -Ls "https://github.com/dustinblackman/phantomized/releases/download/2.1.1/dockerized-phantomjs.tar.gz" | tar xz -C / && \
-  npm install --unsafe-perm --loglevel error && \
+  npm install yarn --global && \
```
sandikbarr: just curious, if Docker still requires npm to work, what's the advantage? Just faster installs?

cavis: I don't think we need it - but this alpine-node image comes with it built in.  And there isn't a yarn package in this version of Alpine, so I can't just `apk add yarn`.  So we need npm to install yarn, but that's it.

I think if we upgraded to the latest alpine node (8.1-ish?) it would have yarn pre-installed.

cavis: Oh wait, you're asking why yarn.  Just to version-lock stuff better than npm, and prevent unexpected breaking upgrades during a docker build.

PR #419, opened 2017-06-22
styleguide image assets

sandikbarr:
gets image loader from the styleguide


cavis:
all looks good now, after styleguide update

PR #421, opened 2017-06-23
move the hero and spinner to the styleguide

sandikbarr:


src/assets/styles/loading.css
```diff
@@ -1,276 +0,0 @@
-.empty-space::before {
```
cavis: Funny - I don't see this `empty-space` class anywhere.  So I guess this file was doing nothing?

sandikbarr: yeah .empty-space was extra but there was some dup styles. We don't need all those browser prefixes for keyframes, so we're good without it.

PR #423, opened 2017-07-06
use fancy form module from styleguide

sandikbarr:


src/app/shared/shared.module.ts
```diff
@@ -40,34 +33,38 @@ import { ImageModule, SpinnerModule } from 'ngx-prx-styleguide';
...
     TextOverflowFadeComponent,
     TextOverflowWebkitLineClampComponent,
+    SpinnerModule,
     WysiwygComponent
   ],
   imports: [
     CommonModule,
     FormsModule,
     RouterModule,
+    DatepickerModule,
```
gcampo88: do these have to be in non-alpha order in order to maintain order of dependencies?

src/app/story/directives/status.component.spec.ts
```diff
@@ -2,7 +2,7 @@ import { cit, create, provide, stubPipe } from '../../../testing';
 import { Router } from '@angular/router';
 import { Angulartics2 } from 'angulartics2';
 import { RouterStub } from '../../../testing/stub.router';
-import { ModalService, ToastrService } from '../../core';
+import { ToastrService, ModalService } from '../../core';
```
gcampo88: are we now deciding to import everything directly from styleguide? if so, i definitely don't think you need to fix every instance here - but maybe we can throw that in a low-priority ticket for later.

src/app/shared/shared.module.ts
sandikbarr: I don't think the order should really make a difference for NgModule. I think what we're doing here is dealing with Angular stuff first then our stuff just for organization, but we didn't really follow that with exports. Probably just lack of setting and following standards.

src/app/story/directives/status.component.spec.ts
sandikbarr: I've got ToastrService coming up behind this one, so I'll address it here with ModalService and make sure Toastr follows the same, but I'll leave component module export/import in shared and core until we prove otherwise.

src/app/series/directives/series-podcast.component.html
```diff
@@ -26,16 +26,16 @@
         that best describe this podcast.
       </div>
       <div class="span-fields">
```
sandikbarr: that does make a difference 👍

sandikbarr:
Updated the publish-fancy-field selectors (oops!) and imported non angular imports directly form the styleguide. I left the core/shared angular stuff and the ordering of imports in the shared module. We can decide about the usefulness of that shared module later (maybe it'll be even more clear what we should do once it's fully gutted of common styleguide components.) And the build passed 🎉, so back to you @gcampo88!

PR #426, opened 2017-07-12
use SVG icons

### LINKED ISSUE #360: timepicker doesn't know about daylight/savings time zones
posting on behalf of @farski, who pointed out: the timepicker does not alter the time zone extension in accordance to daylight/savings time. (e.g.: if I set a release date for an episode in February, it should be HH:MM EST, but if I set a release date for July, it should be HH:MM EDT)

### LINKED ISSUE #424: use timepicker and datepicker from styleguide exclusively
basic.component.ts still uses the old timepicker and datepicker from within publish. update to use styleguide.  https://github.com/PRX/publish.prx.org/blob/master/src/app/story/directives/basic.component.ts#L58

sandikbarr:
also fixes #360 and fully uses the prx-datepicker and fixes #424 

Dependent on https://github.com/PRX/styleguide.prx.org/pull/48


src/app/shared/date/datepicker.component.ts
```diff
@@ -1,72 +0,0 @@
-import { Component, Input, Output, EventEmitter, AfterViewInit, ViewChild, ElementRef } from '@angular/core';
-import * as Pikaday from 'pikaday';
```
cavis: I _think_ you can also clean pikaday out of `package.json` now, right?  But not moment, since that's used outside of the datepicker.

sandikbarr:
Back to you @cavis:
* The big "+" card on the dashboard has something odd going on with the hover state.
  * Fixed in both styleguide by providing hover state for icons inside `a` and `button` and doing some cleanup on display: table/table-cell that wasn't really being fully used
On the search page, the "X" next to the filter-by-series dropdown got bigger and isn't vertically aligned anymore
* Spacing of footer social icons changed (though I don't think I really care about this one)
  * Fixed in the styleguide. The font icons had some margin built in with fontello, which I would prefer not to continue with since we're adding margins and padding elsewhere like on `button` and `.btn-icon`
* Calendar icon on datepicker got bigger (i don't mind this - may even look better)
  * yeah I prefer this too. It looked this way in the styleguide already (and on the episode downloads where we were actually showing two icons) Our standard icon size seems to b 20x20, so this brings it inline with other icons.
* Delete "X" icon in Series -> Audio Templates is off center
  * Fixed centering. This icon also got bigger, but 20x20 is our standard icon size so I prefer it

cavis:
looks good to me!

Issue #422, opened 2017-06-28
show frontend errors if validations fail on backend


sandikbarr:
For series template total length, I entered 0:02:00 for the minimum and nothing for the maximum. I did not get any feedback from the UI that this was not valid. However, when saving the series template, there was a 500 error on the PUT to https://cms-staging.prx.tech/api/v1/audio_version_templates/416 with the message `Validation failed: Length minimum must be less than or equal to 0, Length maximum must be greater than or equal to 119`

While that error message doesn't exactly make sense and is not accurate for what is actually passing validation, front end validation should match what is allowed on the back end.

gcampo88:
I think my vote here would be to alter this on the back end, since it's entirely reasonable to expect the user to set the length min and not the max. CMS ticket to handle: https://github.com/PRX/cms.prx.org/issues/322

But, we can leave this ticket here as one to show frontend errors if validations fail on backend?

Issue #368, opened 2017-02-27
New CMS version/story validation messages


cavis:
Per PRX/cms.prx.org#255, when audio is invalid, the status will propagate up to the audio-version and the story.

Need to display these version/story status messages to the user.  Should only show them when Publish thinks everything is good (no invalids) on that model.  Probably (1) below the audio version, and (2) in the story-status box.

They should have the same effect as a local invalid - prevent saving/publishing the story until they've been resolved.

gcampo88:
v similar to #422

PR #438, opened 2017-08-10
Validate non-matching audio files

### LINKED ISSUE #367: New CMS file validation statuses
Per PRX/cms.prx.org#255...

There are some new audio/image statuses that Publish needs to recognize, and display correctly.  And if set, should display `status_message` to the user, vs some sort of hardcoded Publish message.

- UPLOADED: a new tmp file was PUT/POSTed, and awaiting processing
- NOTFOUND / FAILED: the lambda couldn't find the file, or failed to process it (check `status_message` PRX/cms.prx.org#262)
- TRANSFORMED: (audio only - and probably won't see this status as it immediately changes to something else) lambda completed successfully
- INVALID: (audio only) template validations failed (check `status_message`)
- COMPLETE: all good

cavis:
Requires some CMS changes that aren't in prod yet.  So don't merge!

- [x] Attempt to validate audio format on the frontend
- [x] If things go terribly wrong, show the backend `statusMessage`
  - NOTE: ended up only showing story/version statuses, and not audio-files.  This _should_ catch any problems on the backend though, since those status messages always trickle up to the story.
  - Was difficult to attempt refresh story/version models when the audio processing ended.  So I just clear the status field on-save.  If anything real unexpected happens, user currently has to refresh/renavigate the page or save something else.  Or else they won't see the error.

I think this should fix #367 #368 and #422.

Ugh.


src/app/upload/upload.component.ts
```diff
@@ -85,6 +85,12 @@ export class UploadComponent implements OnInit, DoCheck {
         this.changedClass = !this.versionUndeletedHaveChanged();
       } else if (this.version.changed(null, false)) {
         this.changedClass = true; // version itself changed, not files
+      } else if (this.version.nonMatchingFiles()) {
```
cavis: This is what I feel worst about.  For DRAFT stories, I wanted to let the user save any audio files, but then show the mismatch-error after processing.  Without this special case, the error wouldn't show up in Version-uploader box (presumably because the version hasn't `changed()`) - but would show up in the Story-status-box, and prevent publishing.

src/app/series/series.component.ts
```diff
@@ -90,6 +90,12 @@ export class SeriesComponent implements OnInit {
       if (wasNew) {
         this.router.navigate(['/series', this.series.id]);
       }
+    },
+    err => {
```
sandikbarr: This goes with that questionable 500 error message handling that I was attempting. There's a styleguide branch dependent on it out there too: https://github.com/PRX/styleguide.prx.org/commit/e429c7ce6ff45ddaccfe0cbbb3ae5831fd5f4061 Do you think we should keep it? It felt pretty wrong with the res['_body'] and the JSON.parse()

src/app/shared/model/invalid/audio.invalid.ts
```diff
@@ -47,6 +47,20 @@ export const VERSION_TEMPLATED = (template?: HalDoc): BaseInvalid => {
         return `total length must be less than ${max} - currently ${got}`;
       }
 
+      // file formats must match
+      let nonMatches = [];
+      let labels = {contenttype: 'content type', channelmode: 'channels'};
+      ['contenttype', 'layer', 'frequency', 'bitrate', 'channelmode'].forEach(fld => {
+        let vals = undeleted.map(f => f[fld]).filter(val => val);
```
sandikbarr: all this is 👏 💯 🥇

src/app/upload/shared/audio-duration.component.ts
```diff
@@ -1,20 +1,87 @@
...
+    if (this.file.frequency) {
+      frequency = `${this.file.frequency / 1000} kHz`;
+    }
+    let bitrate = '<i>Unknown</i>';
+    if (this.file.bitrate) {
+      bitrate = `${this.file.bitrate / 1000} kb/s`;
+    }
+    let channels = this.file.channelmode || '<i>Unknown</i>';
+    return this.infoStyle + `
+      <dl>
```
sandikbarr: 😎 ha! this works :) I like how it looks and where you decided to put the info icon with the filename.

src/app/upload/upload.component.ts
sandikbarr: Yeah... at least the episode can't be published and there is feedback.

src/app/upload/shared/audio-duration.component.css
```diff
@@ -0,0 +1,11 @@
...
+  background: transparent url('/assets/images/info-circle.svg') 50% no-repeat;
+  background-size: cover;
+  height: 18px;
+  width: 18px;
+  margin: 0;
+  vertical-align: text-bottom;
+}
+button:hover, button:focus {
+  background-image: url('/assets/images/info-circle-dark.svg')
+}
```
sandikbarr: I woulda had you put this in the styleguide but I'm gonna go ahead an look the other way on it for now. ;)

Issue #431, opened 2017-07-24
Editable Public Feed Url


cavis:
There are some cases where it's necessary to edit the "Public Feed Url" (the atom self link in the feed) without entering a new feed url.

Should either:

1. Make editable all the time, with a whole ton of warnings
2. Make editable until feed has one episode

I'm sort of leaning towards (1), since you could have a bunch of episodes in a non-launched feed, and still be messing around with the public url.

PR #439, opened 2017-08-16
Editable public feed

gcampo88:
In the interest of soliciting feedback, here's one approach to #431: 
- allow edits to public feed URL
- show a checkbox users can check to set the new feed URL to this as well
- lots of messaging that if the feed is live and the public URL changes, the new feed URL should be set as well to redirect subscribers; plus lots of linking to Apple docs on how to handle source feed change.

### DIFF
--- a/src/app/series/directives/series-podcast.component.css
+++ b/src/app/series/directives/series-podcast.component.css
@@ -3,24 +3,9 @@
   width: 250px;
   margin-right: 40px;
 }
-.prx-feed-url input, .feed-url input {
-  padding: 10px 8px;
-  color: #939393;
-  margin-bottom: 4px;
-}
 .prx-feed-url input {
+  padding: 10px 8px;
+  margin-bottom: 4px;
+  color: #939393;
   width: 100%;
-}
-.feed-url input {
-  min-width: 400px;
-}
-.feed-url button, .feed-url .button {
-  padding-top: 9px;
-  padding-bottom: 8px;
-}
-.feed-url button {
-  background-color: #f59f51;
-}
-.feed-url button:hover, .feed-url button:focus {
-  background-color: #e28b42;
 }
--- a/src/app/series/directives/series-podcast.component.css
+++ b/src/app/series/directives/series-podcast.component.css
@@ -3,9 +3,23 @@
   width: 250px;
   margin-right: 40px;
 }
-.prx-feed-url input {
+.prx-feed-url input, .feed-url input {
   padding: 10px 8px;
   margin-bottom: 4px;
-  color: #939393;
+}
+.prx-feed-url input {
   width: 100%;
+}
+.feed-url input {
+  min-width: 400px;
+}
+.feed-url button, .feed-url .button {
+  padding-top: 9px;
+  padding-bottom: 8px;
+}
+.feed-url button {
+  background-color: #f59f51;
+}
+.feed-url button:hover, .feed-url button:focus {
+  background-color: #e28b42;
 }
--- a/src/app/series/directives/series-podcast.component.css
+++ b/src/app/series/directives/series-podcast.component.css
@@ -10,6 +10,9 @@
 .prx-feed-url input {
   width: 100%;
 }
+input.changed {
+  outline: 5px auto #f09b4c;
+}
 .feed-url input {
   min-width: 400px;
 }
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -119,7 +119,7 @@
 
     <prx-fancy-field *ngIf="podcast?.hasPublicFeed" label="Public Feed Url" textinput [model]="podcast" name="publicFeedUrl" [advancedConfirm]="publicFeedChangeConfirm">
       <div class="fancy-hint">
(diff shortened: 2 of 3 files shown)


debenedictis:
I think the phrasing "once your feed is live" could be ambiguous.  A feed could be "live" and yet not have any subscribers.  That happens because the PRX Publishing Platform feed needs to exist before creating the FeedBurner feed.  And, the FeedBurner can exist for a period of time before it is published/announced.

Alternate language could say something like "once you have subscribers" or "once your feed is public" or something else.  Or, maybe I'm splitting hairs.

gcampo88:
that's good feedback @debenedictis! i like "once you have subscribers"

src/app/series/directives/series-podcast.component.html
```diff
@@ -117,21 +117,20 @@
...
 
-    <prx-fancy-field label="Public Feed Url" class="feed-url">
-      <div *ngIf="podcast?.hasPublicFeed">
-        <div class="fancy-hint">
-          The public URL for your podcast feed. Feel free to share this URL with listeners. If you need to alter this URL, please
-          do so safely by using the New Feed Url field in the Advanced settings.
-        </div>
-        <input type="text" readonly [ngModel]="podcast?.publicFeedUrl" name="publicFeedUrl" #pubFeed/>
-        <button [publishCopyInput]="pubFeed">Copy</button>
-        <a class="button" target="_blank" rel="noopener" [href]="podcast?.publicFeedUrl">Open Link</a>
```
sandikbarr: No more Copy or Open Link buttons?

src/app/series/directives/series-podcast.component.css
```diff
@@ -3,24 +3,9 @@
...
   margin-right: 40px;
 }
-.prx-feed-url input, .feed-url input {
+.prx-feed-url input {
   padding: 10px 8px;
-  color: #939393;
   margin-bottom: 4px;
-}
-.prx-feed-url input {
+  color: #939393;
```
sandikbarr: Redundant `color: #939393`. This isn't really your doing, but this color on input is (and has been) specified in the global CSS.

kookster:
I wonder if "new" podcasts are always new - they may just be new to us, and so it would be better to always show the new feed checkbox?

kookster:
another question - I think a common action taken with the public feed url is to add it to directories, like itunes / stitcher / etc. Maybe the language should be talking about sharing this with directories as well / instead of listeners? Could this be the right place to direct people to those directories to add or update a feed?

kookster:
This is vague feedback, but some of the language reads as if changing the public feed url actually changes the podcast's public feed url - it doesn't, this is just where the user tells our system what the public feed url is that they have set up in DNS and feedburner and have or will put it into itunes and their website.

kookster:
yeah, I kinda feel like the checkbox could always be there - not sure we need to show it conditionally, and not being able to see it sometimes could lead to a 'surprise'

src/app/series/directives/series-podcast.component.html
gcampo88: looked a little busy with the checkbox plus the two buttons -- will reinstate and we can see!

gcampo88: oh, and it also is a little tricky  to add advance confirm into the mix -- doesn't play terribly well with publishCopyInput, which wants to have a manually added input field instead of the one given by the publish fancy field (which is the input field that the advance confirm watches for blurring)

sandikbarr: Advanced Confirm can also be user indepedently. From styleguide, for reference: `<input
              prxAdvancedConfirm="That's a risky move" 
              [prxModel]="model" 
              prxName="foo" 
              (ngModelChange)="model.set('foo', $event)"
              [ngModel]="model.foo">`

src/app/series/directives/series-podcast.component.css
gcampo88: 👍

src/app/series/directives/series-podcast.component.html
gcampo88: 🎉 👍

sandikbarr: Unrelated but I stumbled onto something weird when testing. My series has this id on staging, so I had just copied that Homepage Link to the Public Feed URL: https://beta.prx.org/series/36742 😳

gcampo88:
okey doke:
- [X] reinstated the copy + open buttons per @sandikbarr 
- [X] made checkbox always there unless it's first time setting up the podcast per @kookster 
- [X] tweaked language about "live" feed per @debenedictis 

@kookster, I think your idea -- of linking to some sort of guide that a) shows users how to submit URLs to directories like stitcher + itunes, and b) tells them what other steps they'll need to take if they have changed their public feed  -- is a good one. Is a guide like that in the works? I know there's been a lot of back and forth about who submits what to where when, not sure if there's documentation we can link to now, or whether we should open a ticket for that to add later?

sandikbarr:
I'm ok with merging this if @kookster is.

src/app/series/directives/series-podcast.component.html
gcampo88: new screenshot: 
image

(any feelings on checkbox above or below input field?)

kookster:
@sandikbarr I'm all good, pls merge

Issue #435, opened 2017-08-10
Release Date time select is goofy


farski:
Sometimes it includes options for both EDT and EST, other times I've seen it with multiple complete sets of EDT times.

PR #444, opened 2017-08-30
uses the timeseries chart

sandikbarr:
and with update to Styleguide will resolve the bug in #435

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "morgan": "^1.8.1",
     "newrelic": "1.38.2",
     "ng2-dragula": "^1.3.1",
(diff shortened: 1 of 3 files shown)


sandikbarr:
Maybe just not noticeable for you. It can be easy to miss when the date it's given to initialize is already in daylight time. If you select a new date, it adds an additional list of 48 options from 12am-11:30pm in daylight time. It sticks out a lot more when initialized with no date or with a standard time because the timezone suffix is different. Anyway, everytime you change the date, a new set of options was being generated and pushed on without clearing it first. Oops!

cavis:
+good tip!  Confirmed that staging currently adds a ton of select options, and this branch doesn't.

Issue #450, opened 2017-09-05
Validate uploaded files based on template


kookster:
If a template is for videos, allow those to be uploaded, do not apply audio validations.

cavis:
Frontend now does this, though it still fails backend validation as expected.  (Other tickets tackle that).

Issue #449, opened 2017-09-05
Pick a template for a story in a series at create


kookster:

cavis:
Deployed!

Issue #448, opened 2017-09-05
Create multiple version templates for a series


kookster:

cavis:
Deploy'd!

PR #456, opened 2017-09-07
Multiple series templates

cavis:
See #448.

- [x] Allow adding multiple templates to a series (new or existing)
- [x] Allow adding "video file" templates to a series
- [x] Don't allow setting duration or multiple-segments for video templates (hardcoded 1-segment, but it only shows up on the story pages)

Also made sense to build in #450:

- [x] Allow uploading other mime types for a video-file-version.  (Uses `<input accept="video/*"/>`)
- [x] Don't validate video uploads on the frontend

And #449:

- [x] Allow selecting multiple templates associated with your Series Podcast Distribution
  - Force at least 1 selection
- [x] Allow selecting/unselecting the audio-versions for a story
  - Force at least 1 version per story
  - NOTE: this makes **no attempt** to enforce picking a version associated with your podcast (selected above on the series page).  This is fine for now, since there's only 1 kind of distribution.  But later may allow users to shoot themselves in the foot.  Which is also fine.
  - ALSO NOTE: this doesn't warn users that changing around your story-audio-versions can impact what audio is in your feed.  I assume if you changed a story from "Version w/3 segments" to "Version w/4 segments" ... it's fairly self explanatory of the consequences.

### DIFF
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -49,7 +49,7 @@
     <prx-fancy-field required label="Podcast Audio" name="versionTemplateUrls"
       [multiselect]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
       <div class="fancy-hint">
-        Select which template(s) should be used with your podcast. If an episodes
+        Select which template(s) should be used with your podcast. If an episode
         of your podcast provides more than one of these, the first one in this
         list will be used.
       </div>
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -47,8 +47,12 @@
     </prx-fancy-field>
 
     <prx-fancy-field required label="Podcast Audio" name="versionTemplateUrl"
-      [select]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
-      <div class="fancy-hint">Select which audio template should be used with your podcast.</div>
+      [multiselect]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
+      <div class="fancy-hint">
+        Select which template(s) should be used with your podcast. If an episodes
+        of your podcast provides more than one of these, the first one in this
+        list will be used.
+      </div>
     </prx-fancy-field>
 
     <prx-fancy-field label="Homepage Link" required textinput [model]="podcast" name="link">
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -27,7 +27,7 @@
       </div>
       <div class="span-fields">
         <prx-fancy-field label="Category" [model]="podcast" name="category"
-          [select]="categories" (change)="setSubCategories()" small=1>
+          [select]="categories" (onChange)="setSubCategories()" small=1>
         </prx-fancy-field>
         <prx-fancy-field *ngIf="subCategories.length" label="Sub-Category"
           [model]="podcast" name="subCategory" [select]="subCategories" small=1>
@@ -122,7 +122,7 @@
         The public URL for your podcast feed. Feel free to share this URL with listeners. If you need to alter this URL once you have subscribers, set the <a target="_blank" rel="noopener" href="this.itunesNewFeedURLDoc">New Feed URL</a> as well.
       </div>
       <p>
-        <input type="checkbox" (change)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
+        <input type="checkbox" (onChange)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
         <label for="matchFeeds">Check to set your podcast's New Feed URL to this URL as well.</label>
       </p>
       <input type="text" [ngModel]="podcast?.publicFeedUrl" name="publicFeedUrl" #pubFeed
@@ -151,7 +151,7 @@
       </div>
     </prx-fancy-field>
 
-    <prx-fancy-field label="Language" [model]="podcast" name="language" [select]="languageOptions">
+    <prx-fancy-field label="Language" [model]="podcast" name="language" [select]="languageOptions" searchable>
       <div class="fancy-hint">Select which language your podcast is in.</div>
     </prx-fancy-field>
 
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -126,7 +126,7 @@
         The public URL for your podcast feed. Feel free to share this URL with listeners. If you need to alter this URL once you have subscribers, set the <a target="_blank" rel="noopener" href="this.itunesNewFeedURLDoc">New Feed URL</a> as well.
       </div>
       <p>
-        <input type="checkbox" (onChange)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
+        <input type="checkbox" (change)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
         <label for="matchFeeds">Check to set your podcast's New Feed URL to this URL as well.</label>
       </p>
       <input type="text" [ngModel]="podcast?.publicFeedUrl" name="publicFeedUrl" #pubFeed
--- a/src/app/series/directives/series-podcast.component.html
+++ b/src/app/series/directives/series-podcast.component.html
@@ -46,7 +46,7 @@
       </div>
     </prx-fancy-field>
 
-    <prx-fancy-field required label="Podcast Audio" name="versionTemplateUrl"
+    <prx-fancy-field required label="Podcast Audio" name="versionTemplateUrls"
       [multiselect]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
       <div class="fancy-hint">
         Select which template(s) should be used with your podcast. If an episodes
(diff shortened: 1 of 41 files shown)


cavis:
If you really want to run local, here's a hack to change any audio-version-template with "video" in its label to have `contentType = "video/mpeg"`.

```
// in audio-version.model.ts:47
this.template['contentType'] = (this.template['label'] || '').match(/video/i) ? 'video/mpeg' : 'audio/mpeg';

// and in audio-version-template.model.ts:68
if (this.label.match(/video/i)) { this.contentType = 'video/mpeg'; }
```

cavis:
Yeah, hold off for now.  This would break some parts of Publish until CMS actually gets the new field.

src/app/series/directives/series-podcast.component.html
```diff
@@ -47,8 +47,12 @@
     </prx-fancy-field>
 
     <prx-fancy-field required label="Podcast Audio" name="versionTemplateUrl"
-      [select]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
-      <div class="fancy-hint">Select which audio template should be used with your podcast.</div>
+      [multiselect]="audioVersionOptions" [model]="distribution" [advancedConfirm]="versionTemplateConfirm">
+      <div class="fancy-hint">
+        Select which template(s) should be used with your podcast. If an episodes
```
gcampo88: an episodes!

src/app/series/directives/series-podcast.component.ts
```diff
@@ -168,12 +168,9 @@ export class SeriesPodcastComponent implements OnDestroy, DoCheck {
 
   get versionTemplateConfirm(): string {
     if (this.distribution && this.audioVersionOptions && this.series.hasStories) {
-      let url = this.distribution.versionTemplateUrl;
-      let match = this.audioVersionOptions.find(opt => opt[1] === url);
-      let name = match ? match[0] : ''; // options are [[display, value]]
       return `
-        Are you sure you want to use <b>${name}</b> as the audio for your podcast?
-        This will change the audio files used in all published episodes of your podcast.
+        Are you sure you want change the template(s) used for your podcast?
```
gcampo88: want to change!

cavis:
Updated to work locally with PRX/cms.prx.org#336.  Should re-re-test this after that hits staging.

src/app/series/directives/series-podcast.component.html
```diff
@@ -27,7 +27,7 @@
       </div>
       <div class="span-fields">
         <prx-fancy-field label="Category" [model]="podcast" name="category"
-          [select]="categories" (change)="setSubCategories()" small=1>
+          [select]="categories" (onChange)="setSubCategories()" small=1>
```
sandikbarr: I think I see what you're doing here to hook this into the fancy field's prx-select rather than the select change event. Angular actually recommends these EventEmitter outputs not be named `onEvent` and instead just `event`. That doesn't seem to work here, but just mentioning it.

src/app/series/directives/series-podcast.component.html
```diff
@@ -122,7 +126,7 @@
         The public URL for your podcast feed. Feel free to share this URL with listeners. If you need to alter this URL once you have subscribers, set the <a target="_blank" rel="noopener" href="this.itunesNewFeedURLDoc">New Feed URL</a> as well.
       </div>
       <p>
-        <input type="checkbox" (change)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
+        <input type="checkbox" (onChange)="setNewFeedToPublicFeed($event)" [checked]="podcast.publicFeedUrl === podcast.newFeedUrl" id="matchFeeds">
```
sandikbarr: input's don't have an `onChange` event, do they?

src/app/story/directives/basic.component.ts
```diff
@@ -100,11 +114,72 @@ export class BasicComponent implements OnDestroy, DoCheck {
...
+
   toggleShowReleaseAt() {
     this.showReleasedAt = !this.showReleasedAt;
     if (this.story.releasedAt) {
       this.story.releasedAt = null;
     }
   }
 
+  loadVersionTemplates() {
+    this.story.getSeriesTemplates().subscribe(tdocs => {
```
sandikbarr: Navigating from the series episode list to a story threw an error `this.story.getSeriesTemplates is not a function` and would be worth checking here

src/app/series/directives/series-podcast.component.html
cavis: Yeah, and for whenever reason, this immediately stopped working when the underlying field went from a real `<select>` to a fake dropdown.  So I'm guessing the `(change)` event was actually coming from the underlying select, not fired by the FancyFieldComponent itself, and that's why it used to work?

cavis: (Maybe we need to stop event propagation in the fancy field component?)

src/app/series/directives/series-podcast.component.html
cavis: Whoops!  You're right.

src/app/story/directives/basic.component.ts
cavis: Yeah, this is the error with the TabService.  The first/initial time this is called, it gets a SeriesModel, which causes all sorts of trouble.

I'll throw in some patch fixes to make sure `story instanceof StoryModel` ... then remove those when we actually fix the styleguide.

sandikbarr: Not sure I'm on the same page with the fix that's needed in the styleguide. TabService isn't very smart, just a model Observable.

src/app/series/directives/series-templates.component.html
```diff
@@ -0,0 +1,55 @@
...
+      specific segment requirements (note: don't include ads here). Every episode using these templates will
+      be checked against them for validity.
+    </div>
+  </prx-fancy-field>
+
+  <ng-container *ngFor="let v of series.versionTemplates">
+    <div *ngIf="!v.isDestroy" class="version">
+      <header>
+        <strong>{{v?.label}}</strong>
+        <button type="button" class="btn-icon icon-cancel grey-dove" (click)="confirmRemoveVersion(v)"></button>
```
sandikbarr: btn-icon should have `aria-label` for accessibility. (I put it in the styleguide, so this is basically a rule now. 😃 )

src/app/shared/model/audio-version-template.model.ts
```diff
@@ -3,37 +3,45 @@ import 'rxjs/add/observable/of';
...
   public lengthMaximum: number = null;
   public fileTemplates: AudioFileTemplateModel[];
+  private newIndex: number = null;
 
-  SETABLE = ['label', 'lengthMinimum', 'lengthMaximum'];
+  SETABLE = ['label', 'contentType', 'lengthMinimum', 'lengthMaximum'];
 
   VALIDATORS = {
     label: [REQUIRED(), LENGTH(1, 255)],
+    contentType: [REQUIRED(), IN(['audio/mpeg', 'audio/mp2', 'video/mpeg', 'video/x-mpeg'])],
```
sandikbarr: This is nitpicky, but since they are used more than once, I'm wishing these strings were constants.

src/app/story/directives/basic.component.ts
cavis: Update: fixed by providing the TabService at the component level, instead of the shared-module.

src/app/shared/model/audio-version.model.ts
```diff
@@ -106,7 +109,7 @@ export class AudioVersionModel extends BaseModel implements HasUpload {
         this.explicit = 'Clean';
         break;
       default:
-        this.explicit = '';
+        this.explicit = undefined;
```
sandikbarr: I thought we didn't use undefined or null on the model fields? I got tripped up on that 9+ months ago with explicit and have never really understood what the issue was with using it, something about messing up change detection?

src/app/series/series.component.ts
```diff
@@ -4,10 +4,11 @@ import { Observable } from 'rxjs/Observable';
 import { Subject } from 'rxjs/Subject';
 
 import { CmsService } from '../core';
-import { ModalService, ToastrService } from 'ngx-prx-styleguide';
+import { ModalService, TabService, ToastrService } from 'ngx-prx-styleguide';
 import { SeriesModel } from '../shared';
 
 @Component({
+  providers: [TabService],
```
sandikbarr: I did put a ticket in the styleguide to document this so it doesn't trip us up down the road: https://github.com/PRX/styleguide.prx.org/issues/56

src/app/story/directives/basic.component.css
```diff
@@ -7,6 +7,20 @@ hr {
   border-bottom: 1px solid #ccc;
   margin: 0 0 30px;
 }
+
+prx-select {
+  position: absolute;
+  top: -4px;
+  right: 0;
+}
+/* TODO: move to styleguide */
```
sandikbarr: was this moved? (The styleguide has an open issue about making global changed and invalid styles BTW https://github.com/PRX/styleguide.prx.org/issues/30) Bigger problem: I can't seem to get that field to work.

src/app/story/directives/basic.component.ts
```diff
@@ -29,9 +29,14 @@ import { TabService } from 'ngx-prx-styleguide';
...
-        <h1 *ngIf="story?.versions?.length === 0">
+        <prx-select *ngIf="versionTemplateOptions" placeholder="Select Templates..."
+          [selected]="versionTemplatesSelected" [options]="versionTemplateOptions"
+          (onSelect)="updateVersions($event)"
+          [class.invalid]="versionsInvalid"
+          [class.changed]="versionsChanged"></prx-select>
+        <prx-spinner *ngIf="!undeletedVersions"></prx-spinner>
+        <publish-upload *ngFor="let v of undeletedVersions" [version]="v" [strict]="strict"></publish-upload>
+        <h1 *ngIf="undeletedVersions?.length === 0">
           You have no audio templates for this episode. How did that happen?
```
sandikbarr: screen shot 2017-09-21 at 9 29 57 am
Should we keep this cheeky message? If yes, we should probably format it as a fancy field `hint`.

cavis: Yeah - it does look a bit funny next to that other hint.  Will change that - and make it slightly less cheeky.  _Slightly_.

Issue #447, opened 2017-09-05
Max-size validation for Series Image


gcampo88:
We just got a 500 error in feeder prod: `Validation failed: ITunes images height must be less than or equal to 3000, ITunes images width must be less than or equal to 3000` for a Series Image that was 3809x3809 px. 
Per iTunes, "Artwork must be a minimum size of 1400 x 1400 pixels and a maximum size of 3000 x 3000 pixels, in JPEG or PNG format, 72 dpi, with appropriate file extensions (.jpg, .png), and in the RGB colorspace. These requirements are different from the standard RSS image tag specifications."

Publish [only validates the min dimensions and squareness of images](https://github.com/PRX/publish.prx.org/blob/master/src/app/shared/image/image-upload.component.ts#L34); add a max-size validation to Series Image to match the backend validation of the same. (Also, check other backend image validations and see if there are any other max-size validations we need to reflect on frontend).

gcampo88:
Feeder validates the iTunes Image (for which we use the Series Image from Publish) like so:
```Ruby
  validates :height, :width, numericality: {
    less_than_or_equal_to: 3000,
    greater_than_or_equal_to: 1400
  }, if: ->(i) { i.height && i.width }

  validates :height, numericality: { equal_to: -> (image) { image.width } }, if: ->(i) { i.height }
```

Currently, Feeder doesn't validate the Episode Image the same way, but perhaps it should (per [Podcasts Connect](https://help.apple.com/itc/podcasts_connect/#/itcb54353390), which has the same min and max dimensions for itunes:image at both channel and item level)

PR #457, opened 2017-09-07
Validate max size for images

gcampo88:
#447 add max-size messaging + validations for series and story images.

### DIFF
--- a/src/app/shared/image/image-upload.component.ts
+++ b/src/app/shared/image/image-upload.component.ts
@@ -38,8 +38,8 @@ export class ImageUploadComponent implements DoCheck {
   @Input() purpose: string;
   @Input() minWidth = 144;
   @Input() minHeight = 144;
-  @Input() maxWidth = 144;
-  @Input() maxHeight = 144;
+  @Input() maxWidth: number;
+  @Input() maxHeight: number;
   @Input() suggestSize: string;
   @Input() square = false;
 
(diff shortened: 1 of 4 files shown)


src/app/shared/image/image-upload.component.ts
```diff
@@ -33,6 +38,8 @@ export class ImageUploadComponent implements DoCheck {
   @Input() purpose: string;
   @Input() minWidth = 144;
   @Input() minHeight = 144;
+  @Input() maxWidth = 144;
+  @Input() maxHeight = 144;
```
sandikbarr: The series thumbnail image isn't working. Can we maybe not have a default max or something? Or maybe just in the validation below check for either suggestSize or min/max?
screen shot 2017-09-07 at 5 52 56 pm

gcampo88: oooh, good catch!

gcampo88:
good catch @sandikbarr. back to you!

src/app/shared/image/image-upload.component.ts
```diff
@@ -126,6 +117,24 @@ export class ImageUploadComponent implements DoCheck {
...
+    if (!this.suggestSize) {
+      if (width < this.minWidth || height < this.minHeight) {
+        this.imgError = `The image provided is only ${width} x ${height} px
+                         but should be at least ${this.minWidth} x ${this.minHeight} px.`;
+        return false;
+      } else if ((this.maxWidth && width > this.maxWidth) || (this.maxHeight && height > this.maxHeight)) {
+        this.imgError = `The image provided is ${width} x ${height} px
+                         but should not exceed ${this.maxWidth} x ${this.maxHeight} px.`;
+        return false;
+      } else if (this.square && width !== height) {
```
sandikbarr: I wondered if we wanted to enforce square without enforcing min or max, but it looks like we're not doing that, so 👍

Issue #328, opened 2017-02-01
Preview embed before episode is published


kookster:
The trick here is that if the episode isn't published, feeder won't provide unpublished audio via the un-auth API.

We would need to have a different url formed to indicate to dovetail to request the audio from feeder with authorization, such as by including the user's current access token as a param in the audio url, and setting that in the player iframe href.

This then requires dovetail changes to use this token to hit feeder for the episode/audio info, and maybe also treat this differently by not caching the stitch result, or counting it for ad metrics: PRX/dovetail.prx.org/#175

PR #459, opened 2017-09-13
Preview unpublished dovetail audio

kookster:
- [x] Use enclosure url when unpublished
- [x] Strip out enclosure prefix
- [x] Add in the token query param
- [x] Tests, what tests?

### DIFF
--- a/src/app/shared/model/feeder-episode.model.ts
+++ b/src/app/shared/model/feeder-episode.model.ts
@@ -12,11 +12,12 @@ export class FeederEpisodeModel extends BaseModel {
 
   // writeable
   SETABLE = ['guid', 'authorName', 'authorEmail', 'episodeUrl', 'summary'];
-  URLS = ['episodeUrl'];
+  URLS = ['episodeUrl', 'enclosureUrl'];
   guid = '';
   authorName = '';
   authorEmail = '';
   episodeUrl = '';
+  enclosureUrl = '';
   summary = '';
 
   VALIDATORS = {
@@ -52,6 +53,9 @@ export class FeederEpisodeModel extends BaseModel {
     this.authorName = author['name'] || '';
     this.authorEmail = author['email'] || '';
     this.summary = this.doc['summary'] || '';
+    if (this.doc['_links'] && this.doc['_links']['enclosure'] && this.doc['_links']['enclosure']['href']) {
+      this.enclosureUrl = this.doc['_links']['enclosure']['href'] || '';
+    }
   }
 
   encode(): {} {
--- a/src/app/shared/model/feeder-episode.model.ts
+++ b/src/app/shared/model/feeder-episode.model.ts
@@ -53,9 +53,7 @@ export class FeederEpisodeModel extends BaseModel {
     this.authorName = author['name'] || '';
     this.authorEmail = author['email'] || '';
     this.summary = this.doc['summary'] || '';
-    if (this.doc['_links'] && this.doc['_links']['enclosure'] && this.doc['_links']['enclosure']['href']) {
-      this.enclosureUrl = this.doc['_links']['enclosure']['href'] || '';
-    }
+    this.enclosureUrl = this.doc.expand('enclosure') || '';
   }
 
   encode(): {} {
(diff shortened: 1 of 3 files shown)


src/app/shared/model/feeder-episode.model.ts
```diff
@@ -52,6 +53,9 @@ export class FeederEpisodeModel extends BaseModel {
     this.authorName = author['name'] || '';
     this.authorEmail = author['email'] || '';
     this.summary = this.doc['summary'] || '';
+    if (this.doc['_links'] && this.doc['_links']['enclosure'] && this.doc['_links']['enclosure']['href']) {
+      this.enclosureUrl = this.doc['_links']['enclosure']['href'] || '';
```
sandikbarr: Should enclosure be a related model? I know it's right there in _links, but we don't access _links, we follow them.

src/app/shared/model/feeder-episode.model.ts
```diff
@@ -52,6 +53,7 @@ export class FeederEpisodeModel extends BaseModel {
     this.authorName = author['name'] || '';
     this.authorEmail = author['email'] || '';
     this.summary = this.doc['summary'] || '';
+    this.enclosureUrl = this.doc.expand('enclosure') || '';
```
sandikbarr: 👍 at least I feel better

src/app/story/directives/player.component.ts
```diff
@@ -46,6 +50,7 @@ export class PlayerComponent implements OnDestroy, DoCheck {
         }
       });
     });
+    auth.token.subscribe(token => this.authToken = token);
```
sandikbarr: I think you'll want to look in app.component.spec.ts for an example of how to mock this token in a spec.

src/app/shared/model/feeder-episode.model.ts
kookster: that is usually true. in this case the (content) type of the link is not to another json-hal resource, it's to the enclosure audio/mpeg. I see about 50/50 folks using _links for non-hal resources in their APIs, there is a `type` attr on the link object to support this way of using it.

src/app/shared/model/feeder-episode.model.ts
kookster: it was a good note - definitely that code was a cry for help.

src/app/story/directives/player.component.ts
kookster: thanks!

kookster:
#328

kookster:
thanks @sandikbarr! I think the test I added kicks the tires at least, see if you think this is good enough (and thanks for the tip, `app.component.spec.ts` was helpful to look at).

src/app/story/directives/player.component.spec.ts
```diff
@@ -64,6 +68,18 @@ describe('PlayerComponent', () => {
     expect(comp.episodeGuid).toEqual('episode1');
   });
 
+  cit('uses available enclosure url for unpublished', (fix, el, comp) => {
+    let dist = series.mock('prx:distribution', {kind: 'podcast', url: 'http://some-where'});
+    dist.mock('http://some-where', {publishedUrl: 'http://published-url', enclosurePrefix: 'http://prefix/'});
+    let storyDists = story.mockItems('prx:distributions', [{kind: 'episode', url: 'http://some-where/episode'}]);
+    storyDists[0].mock('http://some-where/episode', {guid: 'episode1', _links: { enclosure: { href: 'http://prefix/some-where/enclosure.mp3'}}});
+    comp.fromFeeder(new StoryModel(series, story), new DistributionModel(series, dist));
```
sandikbarr: Looks like maybe you had to call these methods directly rather than let the component do the thing, but I'm satisfied with this test. It might have worked to call ngDoCheck as well. I've had to do that with ngOnChanges because the component doesn't seem to go through the on change lifecycle hooks with tests.

kookster: I truly have no idea if this is best, copied the style of the other tests for this part of the code.

Issue #467, opened 2017-10-05
List page should highlight invalid episodes 


gcampo88:
the Series --> Episodes in Series list should display invalid episodes differently. Right now there's no way for producers to see at a glance which episodes are invalid. Would be very helpful for producers whose feeds have just been imported, or who are altering audio version templates. (e.g. of these, several are invalid, but no way to tell which without clicking through every ep)

![image](https://user-images.githubusercontent.com/16007382/31232997-7034e6e6-a9b9-11e7-8d3b-1af9bedb4c29.png)

PR #470, opened 2017-10-10
Highlight invalid episodes on series feed page

gcampo88:
#467 

in action: 
image

### DIFF
--- a/src/app/series/directives/series-feed.component.css
+++ b/src/app/series/directives/series-feed.component.css
@@ -35,10 +35,11 @@ li p {
 .futurePublic {
   color: #ffa500;
 }
-
-
 .extra {
   text-align: center;
   margin: 20px;
   font-weight: 400;
+}
+.invalid {
+  color: #e32;
 }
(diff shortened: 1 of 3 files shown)


src/app/series/directives/series-feed.component.css
```diff
@@ -35,10 +35,11 @@ li p {
...
 }
-
-
 .extra {
   text-align: center;
   margin: 20px;
   font-weight: 400;
 }
+.invalid {
+  color: #e32;
```
sandikbarr: Are we using this red color anywhere else? Our non-emphasis red is #ee4d3e? Does that not look good?

gcampo88:
this is the same red as in the status box for an invalid ep:
image

 happy to change though!

src/app/series/directives/series-feed.component.css
sandikbarr: 😎 Cool, if we are using it for text elsewhere, let's keep as is! (Invalid color/style probably belongs in the styleguide, but that's something we can evaluate another day with this ticket https://github.com/PRX/styleguide.prx.org/issues/30.)

PR #480, opened 2017-10-27
Use zoom param

cavis:
Use `?zoom=` to speed up dashboard, search, and series-episode-list.


src/app/shared/model/series.model.ts
```diff
@@ -45,11 +45,13 @@ export class SeriesModel extends BaseModel implements HasUpload {
       this.set('accountId', account.id, true);
     }
     this.hasStories = series ? series.count('prx:stories') > 0 : false;
-    this.loadRelated('versionTemplates').subscribe(() => {
```
cavis: This was causing any `new SeriesModel()` to force-follow the audio version templates.  Even if you passed `loadRelated = false` in the constructor.

Issue #496, opened 2018-01-18
Strip out spaces from URLs before save 


gcampo88:
(just had an issue where the enclosure prefix url had a trailing space)

gcampo88:
I think we can call this handled with #502

PR #502, opened 2018-02-14
Upstream fix for url validator

cavis:
Doesn't actually solve #496, but at least fixes the URL-validations so you can't save your messed up URL.

Still has the issue where "strange things" can happen if you type in a leading space, or delete part of "http://".


gcampo88:
any issue w/ merging this? i remember some slack back + forth about computer getting very warm?

cavis:
yep, that's fixed.  plus, a warm computer is a feature.  makes the electrons go faster.

PR #505, opened 2018-02-23
Use published_released_at virtual column sorting

pkarman:


pkarman:
@cavis I'm not sure how to best test this?

cavis:
Not a real useful way to unit test it, mocking out CMS.  I'd probably just QA it - create a few stories within a series, with various null published_at/released_at dates.

src/app/story/directives/status.component.ts
```diff
@@ -87,6 +91,7 @@ export class StoryStatusComponent implements DoCheck {
     if (this.story) {
       this.setStatus();
       this.isPublished = this.story.publishedAt ? true : false;
+      this.isReleased = this.story.releasedAt ? true : false;
```
cavis: Nitpick: only show this when not published.  Otherwise, published episodes get this:

img

And publishing (you clicked publish, but set a releaseAt in the future) get this:

img

src/app/story/directives/status.component.ts
```diff
@@ -26,6 +26,9 @@ import { StoryModel } from '../../shared';
       <dt *ngIf="isPublished && !isScheduled">Published</dt>
       <dd *ngIf="isPublished"><p>{{story.publishedAt | date:"short"}}</p></dd>
 
+      <dt *ngIf="isReleased">Released</dt>
```
cavis: Maybe use "Release" instead of "Released", since any episode showing this will not actually be published?  (Also matches the "Release Date" language at the bottom of the form).

src/app/story/directives/status.component.ts
pkarman: good points. fixed in bcbdb44

screen shot 2018-03-01 at 2 57 53 pm

screen shot 2018-03-01 at 2 58 24 pm

PR #507, opened 2018-03-30
Permission denied route

pkarman:
Adds a route for `/permission-denied` when the styleguide AuthService responds with an invalid_scope error.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,8 +1,8 @@
 <prx-auth [host]="authHost" [client]="authClient"></prx-auth>
 <prx-header>
   <prx-navitem *ngIf="loggedIn" route="/" text="Home"></prx-navitem>
-  <prx-navitem *ngIf="loggedIn" route="/series/new" text="New Series"></prx-navitem>
-  <prx-navitem *ngIf="loggedIn" route="/search" text="Search"></prx-navitem>
+  <prx-navitem *ngIf="authorized" route="/series/new" text="New Series"></prx-navitem>
+  <prx-navitem *ngIf="authorized" route="/search" text="Search"></prx-navitem>
   <prx-navuser *ngIf="loggedIn" [userName]="userName">
     <prx-spinner class="user-loading" inverse=true></prx-spinner>
     <prx-image *ngIf="userImageDoc" class="user-loaded" [imageDoc]="userImageDoc"></prx-image>
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -3,7 +3,7 @@
   <prx-navitem *ngIf="loggedIn" route="/" text="Home"></prx-navitem>
   <prx-navitem *ngIf="authorized" route="/series/new" text="New Series"></prx-navitem>
   <prx-navitem *ngIf="authorized" route="/search" text="Search"></prx-navitem>
-  <prx-navuser *ngIf="loggedIn" [userName]="userName">
+  <prx-navuser *ngIf="loggedIn" [userinfo]="userinfo">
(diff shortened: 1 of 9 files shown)


pkarman:
note to self: CI will fail until the styleguide module is updated and released and indicated here in package.json

src/app/app.routing.ts
```diff
@@ -3,11 +3,13 @@ import { Routes, RouterModule } from '@angular/router';
 
 import { AuthGuard, UnauthGuard } from 'ngx-prx-styleguide';
 import { ErrorComponent } from './error';
+import { AuthorizationComponent } from './authorization/authorization.component';
 import { HomeComponent, HomeSeriesComponent, HomeStoryComponent } from './home';
 import { LoginComponent } from './login/login.component';
 
 export const routes: Routes = [
   { path: '_error_',   component: ErrorComponent },
+  { path: 'permission-denied', component: AuthorizationComponent },
```
cavis: Nitpick: could give this a canActivate, similar to `login`, which redirects you back to the root when not applicable.  But in real life ... not likely to happen.

src/app/app.component.html
```diff
@@ -1,9 +1,9 @@
 <prx-auth [host]="authHost" [client]="authClient"></prx-auth>
 <prx-header>
   <prx-navitem *ngIf="loggedIn" route="/" text="Home"></prx-navitem>
-  <prx-navitem *ngIf="loggedIn" route="/series/new" text="New Series"></prx-navitem>
-  <prx-navitem *ngIf="loggedIn" route="/search" text="Search"></prx-navitem>
-  <prx-navuser *ngIf="loggedIn" [userName]="userName">
+  <prx-navitem *ngIf="authorized" route="/series/new" text="New Series"></prx-navitem>
+  <prx-navitem *ngIf="authorized" route="/search" text="Search"></prx-navitem>
+  <prx-navuser *ngIf="loggedIn" [userinfo]="userinfo">
```
cavis: Seeing this for reals now ... I'm noting that my display went from "Ryan Cavis" to "ryancavis".  Probably because my preferred_username is `ryancavis`?  That sure sounds like the right thing to do, given that column name.  But if everyone has their lowercase-username in that column, I guess I'd prefer to use the id `name:` over `preferred_username`.  Principle of least surprise!

cavis: I'm also not a huge fan of how the dropdown is unnecessarily wider than my username.  There seems to be some sort of min-width 200px on it?  Causing that padding on the left/right sides.

image

pkarman: I think I got the `min-width` from how the corporate site widget works. I could get rid of it altogether (which I suspect might do weird things), or make it smaller, like `150`. Thoughts?

pkarman:
@cavis the `preferred_username` bit is in the styleguide. I could revert to `name` there if you like.

pkarman:
All the desired changes are in the styleguide. See https://github.com/PRX/styleguide.prx.org/pull/97

src/app/app.routing.ts
cavis: Well that's odd!  Yeah, don't worry about it then.

src/app/app.component.html
cavis: Seems like you got rid of it altogether!  Which is fine by me, since that list will always contain `PRX Exchange` for every user.

pkarman:
screen shot 2018-04-16 at 4 34 27 pm

pkarman:
Thanks @cavis -- I bumped the styleguide version (2x!) and expect the CI to turn 💚

Issue #550, opened 2018-10-01
List-all-episodes page timeout


cavis:
Timeouts trying to list > 3K stories for a series.  Ideas:

- Request in batches
- New CMS endpoint to return very-very _very_ minimal data (title + publishedAt + id)

PR #551, opened 2018-10-09
Use some paging while loading full episode list

cavis:
The easier fix for #550.

Still important that this page contain _all_ episodes/titles, for searchability.  But to avoid overwhelming CMS, load the episodes 50 at a time.

### DIFF
--- a/src/app/series/directives/series-feed.component.spec.ts
+++ b/src/app/series/directives/series-feed.component.spec.ts
@@ -1,4 +1,4 @@
-import { cit, create, provide, stubPipe } from '../../../testing';
+import { cit, create, provide, stubPipe, cms } from '../../../testing';
 import { SeriesFeedComponent } from './series-feed.component';
 import { TabService } from 'ngx-prx-styleguide';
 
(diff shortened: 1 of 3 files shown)


src/app/series/directives/series-feed.component.spec.ts
```diff
@@ -23,7 +23,7 @@ describe('SeriesFeedComponent', () => {
     comp.isLoaded = true;
     fix.detectChanges();
 
-    expect(el).toContainText('You have no published episodes');
+    expect(el).toContainText('You have no episodes');
```
cavis: This text wasn't quite accurate, since this page shows published + released + draft episodes.

src/app/series/directives/series-feed.component.ts
```diff
@@ -66,57 +69,84 @@ import { SeriesModel, StoryModel } from '../../shared';
...
   isLoaded = false;
-  noStories: boolean;
-  series: SeriesModel;
-  publicStories: StoryModel[] = [];
-  futurePublicStories: StoryModel[] = [];
-  privateStories: StoryModel[] = [];
+  isPaging = false;
+  noStories = false;
+  series: HalDoc;
+  publicStories: HalDoc[] = [];
```
cavis: Simplified from the Models down to regular HalDocs (which tend to work better/faster when things are read-only).

PR #565, opened 2018-12-05
don't insert filename in header

svevang:
Fixes uploads for utf8 encoded filenames. Since you can't use utf8 in headers, and this is not really used in the upload, this PR removes the content-disposition header.


cavis:
So many reviews!

PR #596, opened 2019-02-23
allows native-validation "mpeg" through as an mp3

sandikbarr:
there may be better ways to fix this


sandikbarr:
I got a chance to satisfy my curiosity on this. It actually doesn't seem related to the id3 image but something about the files where aurora doesn't ever see the asset format or metadata. I ffprobe'd a few different files (some with images that were working fine) to see if I could figure out anything obvious but it's not jumping out at me. Ticket #597

PR #600, opened 2019-02-28
Limit max audio template selection to 1

pstjean:
Rather than moving to a single select, which changes the underlying datatype to a `number` rather than an `number[]`, I've limited max selection to one. Since we're still planning to re-enable multiple templates in the future, it doesn't make sense to change the underlying Angular model.

### DIFF
--- a/src/app/story/directives/basic.component.ts
+++ b/src/app/story/directives/basic.component.ts
@@ -38,18 +38,18 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
       <prx-fancy-field required label="Audio Files">
         
         <prx-select
-          *ngIf="versionTemplateOptions" class="file-select"  placeholder="Select Templates..."
+          *ngIf="versionTemplateOptions" class="file-select" placeholder="Select Template..."
           [selected]="versionTemplatesSelected" [options]="versionTemplateOptions"
-          (onSelect)="updateVersions($event)" [maxSelectedItems]="1"
-          [class.invalid]="versionsInvalid"
-          [class.changed]="versionsChanged"></prx-select>
+          (onSelect)="updateVersions($event)" [maxSelectedItems]="1" [closeOnSelect]="true"
+          [class.invalid]="versionsInvalid" [class.changed]="versionsChanged">
+        </prx-select>
         <prx-spinner *ngIf="!undeletedVersions"></prx-spinner>
         <publish-upload *ngFor="let v of undeletedVersions" [version]="v" [strict]="strict"></publish-upload>
         <div *ngIf="undeletedVersions?.length === 0" class="fancy-hint">
-          Pick at least one version of your audio files to upload for this episode.
+          Pick a version of your audio files to upload for this episode.
         </div>
         <div *ngIf="undeletedVersions?.length > 0" class="fancy-hint">
-          Clear your selection to select another template
+          Clear your selection to select another template.
         </div>
       </prx-fancy-field>
 
--- a/src/app/story/directives/basic.component.ts
+++ b/src/app/story/directives/basic.component.ts
@@ -36,15 +36,20 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
       <hr/>
 
       <prx-fancy-field required label="Audio Files">
-        <prx-select *ngIf="versionTemplateOptions" class="file-select"  placeholder="Select Templates..."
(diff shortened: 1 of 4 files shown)


pstjean:
Worth noting: this doesn't break an existing episode with more than one audio template selected (I don't know whether those actually exist in production data, but I tested it in staging just in case). The validation is happening purely on the select component itself, not the model, and fires only when the field is changed. If the user clears out the field, however, they can't get back to the multiple-templates-selected state.

src/app/story/directives/basic.component.ts
```diff
@@ -36,16 +36,21 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
...
-          (onSelect)="updateVersions($event)"
+          (onSelect)="updateVersions($event)" [maxSelectedItems]="1"
           [class.invalid]="versionsInvalid"
           [class.changed]="versionsChanged"></prx-select>
         <prx-spinner *ngIf="!undeletedVersions"></prx-spinner>
         <publish-upload *ngFor="let v of undeletedVersions" [version]="v" [strict]="strict"></publish-upload>
         <div *ngIf="undeletedVersions?.length === 0" class="fancy-hint">
           Pick at least one version of your audio files to upload for this episode.
         </div>
+        <div *ngIf="undeletedVersions?.length > 0" class="fancy-hint">
```
cavis: How does one get into this state?  The dropdown won't let me select > 1 in the first place.  (Or maybe this is just a belt-and-suspenders thing).

In any case ... I find it confusing why I can't click the expand-dropdown at all when I have a template selected.  Like there's no UI feedback on that.  Just the "Clear your selection" message.

src/app/story/directives/basic.component.ts
```diff
@@ -36,16 +36,21 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
       <hr/>
 
       <prx-fancy-field required label="Audio Files">
-        <prx-select *ngIf="versionTemplateOptions" class="file-select"  placeholder="Select Templates..."
+        
+        <prx-select
```
cavis: Pre-existing styleguide thing ... but what's with the 2 different delete X's on the dropdown?

src/app/story/directives/basic.component.ts
```diff
@@ -36,16 +36,21 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
...
+          *ngIf="versionTemplateOptions" class="file-select"  placeholder="Select Templates..."
           [selected]="versionTemplatesSelected" [options]="versionTemplateOptions"
-          (onSelect)="updateVersions($event)"
+          (onSelect)="updateVersions($event)" [maxSelectedItems]="1"
           [class.invalid]="versionsInvalid"
           [class.changed]="versionsChanged"></prx-select>
         <prx-spinner *ngIf="!undeletedVersions"></prx-spinner>
         <publish-upload *ngFor="let v of undeletedVersions" [version]="v" [strict]="strict"></publish-upload>
         <div *ngIf="undeletedVersions?.length === 0" class="fancy-hint">
           Pick at least one version of your audio files to upload for this episode.
```
cavis: Should probably update this line.

src/app/story/directives/basic.component.ts
cavis: (this is the UX i find non-intuitive)

pstjean: I agree. I can work on better feedback for this.

src/app/story/directives/basic.component.ts
pstjean: Because switching from multiple select to single select changes the underlying data source from an `any[]` to an `any` this is a multiple select with a maximum of one item selectable. The X on the right clears all items, while the X on the left clears that specific item. In this particular case, there's no difference.

Awkward, not sure it's worth working around? We could hide the right-hand X for a `prx-select` with `maxSelectedItems === 1`

brandonhundt: +1 for this 
> We could hide the right-hand X for a prx-select with maxSelectedItems === 1

pstjean: Not even sure where to put this, as most changes are in Styleguide, but this is what I ended up with:

image

brandonhundt: Great. I like it!

PR #607, opened 2019-03-01
Replace story Category text input w/ prx-tags

rpeterman-gp:
- adds _Ad-Free_ toggle to Categories field
- refactor story model to handle tags as array always
- remove test related to tags parsing

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "langmap": "0.0.14",
     "moment": "^2.18.1",
     "ng2-dragula": "^1.3.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "langmap": "0.0.14",
     "moment": "^2.18.1",
     "ng2-dragula": "^1.3.1",
-    "ngx-prx-styleguide": "2.1.1",
+    "ngx-prx-styleguide": "^2.2.0",
```
sandikbarr: We've been pinning this to the exact styleguide version.

PR #611, opened 2019-03-11
Add production notes to stories

pstjean:
Feedback on the label for this field is welcome. A bit tricky to communicate what it's for.


pstjean:
Don't know how I missed that test. Fixed.

PR #613, opened 2019-03-14
Add sticky positioning to story publish box

rpeterman-gp:
Edit any episode and scroll page. Publish box and Delete button should stay in view.

### DIFF
--- a/src/app/shared/sticky/sticky.directive.ts
+++ b/src/app/shared/sticky/sticky.directive.ts
@@ -1,0 +1,15 @@
+import { Directive, ElementRef } from '@angular/core';
+import * as Stickyfill from 'stickyfilljs';
+
+@Directive({
+  selector: '.sticky-container'
+})
+export class StickyDirective {
+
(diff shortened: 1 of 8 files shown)


src/app/shared/sticky/sticky.directive.ts
```diff
@@ -0,0 +1,15 @@
+import { Directive, ElementRef } from '@angular/core';
+import * as Stickyfill from 'stickyfilljs';
+
+@Directive({
+  selector: '.sticky-container'
```
sandikbarr: Our other directives user an attribute selector, but if it's an established standard, it's an unspoken one, just what's been done. I see you're also using the class with CSS, so maybe that was your thinking. I'm not at all picky about it, just pointing it out as maybe something we should standardize on.

rpeterman-gp: Normally I would use an attribute, but in this case it is the use of the class that needs pollyfilled, which is all this directive does, and is only needed until we don't support <= IE 11.

PR #614, opened 2019-03-15
production calendar dashboard/home page

sandikbarr:
implements the new home page, lives in a separate route until "release"


src/app/dashboard/dashboard-story.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    } else if (!this.story.publishedAt) {
+      this.status = 'draft';
+    } else if (!this.story.isPublished()) {
+      this.status = 'scheduled';
+    } else {
+      this.status = 'published';
+    }
+  }
+
+  loadData() {
```
pstjean: This method name gives the impression that this data is being fetched from a remote source.

sandikbarr: yeah... honestly there are some weird conventions and old bad habits in this code that I'm refactoring a little bit here. Stay tuned for an update! All this came from the existing home component stuff since that's where I initially started building this (instead of building new components from the start.)

sandikbarr:
@pstjean I did some refactoring on this to get rid of nested subscriptions. We're unfortunately doing that in a lot of places in publish, and that's what I found when I went poking around to figure out what I needed to do to get the podcast and episode distributions from feeder. Back to you!

src/app/dashboard/dashboard.component.ts
```diff
@@ -0,0 +1,42 @@
...
+  series: SeriesModel[];
+
+  constructor(private cms: CmsService) {}
+
+  ngOnInit() {
+    this.isLoaded = false;
+    this.cms.auth.pipe(
+      mergeMap((auth: HalDoc) => {
+        this.auth = auth;
+        return auth.follow('prx:default-account');
```
sandikbarr: TODO: defaultAccount for stories without series(pretty sure?) but the series has a `prx:account` and not just the default

sandikbarr: fixed

Issue #624, opened 2019-04-02
series performance


sandikbarr:
* Should get feeder ids from url
100+s of requests of the new home page to feeder just to get the podcast and feeder ids. Instead of following the distribution links, "hack" the id from the url so it doesn't have to make all those requests.
* Don't default loadRelated on the series
https://github.com/PRX/publish.prx.org/blob/master/src/app/dashboard/dashboard.component.ts#L47
Those should maybe have a 3rd param of `false` (loadRelated defaults to `true`)
The home page needs the distributions but not the audio templates.
* Don't load the series account
you can just pass `null` as the first parameter

cavis:
Per the "feeder ids from distribution url" ... I can see 2 ways to do that:

1. Be specific - add the logic to the DistributionModel and StoryDistributionModel
2. Be generic - add it to HalDoc in styleguide.  Some sort of a `doc.guessIdFromLink(linkName)`

PR #626, opened 2019-04-03
improves performance by eliminating feeder requests for ids

sandikbarr:
addresses #624 

I started to go the generic route with the guess id in the haldoc, but then I realized these actually weren't even hal links, they are url properties on a resource that we're treating like a link, and it was really looking like a static function in haldoc that didn't feel right, so I chose to go the specific route in the distribution models. Hopefully that's more of a guiding factor for future developers as well as guessing a link is not the recommended path (except in special cases I suppose.)

### DIFF
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -1,6 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { Observable, forkJoin } from 'rxjs';
-import { mergeMap, map } from 'rxjs/operators';
+import { mergeMap } from 'rxjs/operators';
 import { CmsService, HalDoc } from '../core';
 import { SeriesModel } from '../shared';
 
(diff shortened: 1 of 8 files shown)


src/app/shared/model/distribution.model.ts
```diff
@@ -86,6 +86,13 @@ export class DistributionModel extends BaseModel {
     return {podcast, versionTemplates};
   }
 
+  get feederPodcastId(): string {
+    if (this.kind === 'podcast' && this.url) {
+      const parts = this.url.match(/\/(.[^\/]*)$/);
```
cavis: This regex is a bit hard to follow.  Maybe more like a `this.url.split('/').pop()` or something?

src/app/dashboard/dashboard.component.ts
```diff
@@ -33,18 +30,11 @@ export class DashboardComponent implements OnInit {
         this.defaultAccount = defaultAccount;
         return this.auth.followItems('prx:series', {filters: 'v4', zoom: 'prx:image'});
```
cavis: I think if this zoom was `prx:image,prx:distributions` that would cut down on 10 more requests.

sandikbarr:
Quality suggestions, thanks! Back to you @cavis

PR #627, opened 2019-04-03
Timezone datepicker from styleguide

pstjean:


package.json
```diff
@@ -83,6 +82,7 @@
     "karma-jasmine": "~1.1.0",
     "karma-jasmine-html-reporter": "^0.2.2",
     "karma-spec-reporter": "0.0.26",
+    "moment": "^2.24.0",
```
cavis: Hmm... why is this just a dev dependency now?  \me reads on!

src/app/story/directives/basic.component.ts
```diff
@@ -95,12 +95,9 @@ import { HalDoc, ModalService, TabService } from 'ngx-prx-styleguide';
         to either delay or back-date its release, select the desired release date and time here.
         Otherwise, the episode will be released immediately once published.
         </div>
-        <prx-datepicker *ngIf="showReleasedAt"
+        <prx-tz-datepicker *ngIf="showReleasedAt"
```
cavis: I notice the timezone part of this has some vertical issues, with the cursor / text / carat not all lining up.  Seems to only happen if you try to change the timezone - it starts out correctly aligned.

image

cavis: I also wish this field allowed null values without being marked as "Invalid Date".  (As the old date/timepickers did).  A bit jarring to click the "Specify date and time" and immediately being hit with the red outline.

package.json
cavis: Huh.  Is this only for that 1 `alerts when unscheduling a future published episode` test?  Could we just convert that to basic js Date() and get rid of the dev dependency entirely?

angular.json
```diff
@@ -21,9 +21,10 @@
               "src/assets",
               "src/favicon.ico",
               {
-                "glob": "*",
-                "input": "node_modules/ngx-prx-styleguide/assets/images",
-                "output": "assets/images/"
+                "glob": "**/*",
+                "input": "node_modules/ngx-prx-styleguide/assets",
```
cavis: I'm seeing 2 requests to `/assets/data/timezones.json` every time the field is rendered.  Should probably get that down to 1.

Also, this might be a prx-ng-serve problem, but that timezone file is set to `Cache-Control: no-cache`.  Which isn't ideal.

src/app/story/directives/basic.component.ts
cavis: Maybe a compromise would be to explicitly set `story.releasedAt = new Date()` whenever that checkbox is checked.  And set it back to null when unchecked.

angular.json
pstjean: All of our assets seem to be served with a `Cache-Control: no-cache` header, not just this particular file. I'll open a ticket in prx-ng-serve.

pstjean: See PRX/prx-ng-serve#4

pstjean:
Styleguide bug fix to address the strange "Discard" behavior: PRX/styleguide.prx.org#175

PR #633, opened 2019-04-16
double bearer workaround

sandikbarr:


src/app/series/production-calendar/production-calendar-series.component.ts
```diff
@@ -167,7 +167,7 @@ export class ProductionCalendarSeriesComponent implements OnInit {
     if (this.firstStoryDate && this.lastStoryDate) {
       let months = [];
       let date = new Date(this.firstStoryDate);
-      while (date.valueOf() < this.lastStoryDate.valueOf()) {
+      while (date.valueOf() <= this.lastStoryDate.valueOf()) {
```
sandikbarr: oops, also saw this didn't quite work with a single date/episode

src/app/series/production-calendar/production-calendar-series.component.ts
```diff
@@ -109,7 +109,7 @@ export class ProductionCalendarSeriesComponent implements OnInit {
       this.firstStoryDate ? of([{publishedAt: this.firstStoryDate}]) : this.loadTopStory('asc'),
       this.lastStoryDate ? of([{publishedAt: this.lastStoryDate}]) : this.loadTopStory('desc'),
       this.series.doc.followItems('prx:stories', {
-        per,
+        per: this.PER_SERIES, // don't use per/count for now to work around halcache sticking on an extra Bearer token
```
sandikbarr: Instead of requesting the per limited by the count of the number of stories as we normally do, per will just be 10. The count number isn't entirely accurate here anyway as it doesnt take filters into account. This works around the issue of the halcache putting in two Bearer tokens because it doesn't duplicate the request for the single "top" story from today forward ascending.

cavis: So weird that this would cause the issue.  Ah well, guess it'll be investigated per PRX/styleguide.prx.org#173.

Issue #608, opened 2019-03-05
per series Dashboard


sandikbarr:
![publish homepage](https://user-images.githubusercontent.com/2250413/53820054-b4741f80-3f30-11e9-8be4-930f96de6522.png)
Per series dashboard view
* [x] lays stories out in a vertical list default ordered by publishedAt || lastUpdated
* [x] removes the episode images
* [ ] Has a link to the episode metrics (requires getting the podcast audio version for the guid) (support needed in metrics to enable direct linking to an episode guid)
* [ ] Allows filtering by publish state (needs support in CMS)
* [x] ~Allows episodes to be sorted by _____?~
* [x] New episode buttons move to each series header

brandonhundt:
@sandikbarr Updated Prototype here: https://dawn-flower-2080.animaapp.io/publishhomepage

PR #634, opened 2019-04-19
production calendar release

sandikbarr:


src/app/dashboard/dashboard-series.component.css
```diff
@@ -3,13 +3,13 @@ header {
   justify-content: space-between;
   align-items: flex-end;
   position: relative;
-  overflow: auto;
+  overflow: hidden;
```
sandikbarr: buttons should be aligned per #608 but were looking like this with long series names
Screen Shot 2019-04-22 at 9 14 56 AM

so I fixed em up
Screen Shot 2019-04-22 at 9 30 20 AM

Issue #598, opened 2019-02-27
Series owner field clear and discard should repopulate owner


sandikbarr:
There is a bug with the series Owner field on the Basic tab that if you clear the Owner form control then click Discard, the existing owner does not repopulate in the field.

rpeterman-gp:
I looked into this, but was unable to get my local env running to test it out. From what I could fine, it looks like the series model is not mapping the `accountId` from the doc in its `decode` method.

Issue #436, opened 2017-08-10
Episode URL is marked as a required field


farski:
The Episode URL field is marked as required (*), even though it doesn't seem to be required. The description for it even says "if…"

sandikbarr:
Such a small change. Low priority but hopefully we'll just pick it up and fix it.

PR #637, opened 2019-04-25
Fixes series owner discard

sandikbarr:
#598 Fixes series owner/accountId discard bug. The series owner field was not getting restored on discard because it's a sort of pseudo field that actually comes from the parent account.

#436 Also removes feeder episode url from required fields

### DIFF
--- a/src/app/shared/model/series.model.ts
+++ b/src/app/shared/model/series.model.ts
@@ -120,6 +120,9 @@ export class SeriesModel extends BaseModel implements HasUpload {
   }
 
   discard(): any {
+    if (this.changed('accountId')) {
+      this.set('accountId', this.original['accountId']);
+    }
     super.discard();
     if (this.isNew) {
       this.defaultVersionTemplate();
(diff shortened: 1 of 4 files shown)


src/app/shared/model/series.model.ts
```diff
@@ -120,6 +120,9 @@ export class SeriesModel extends BaseModel implements HasUpload {
   }
 
   discard(): any {
+    if (this.changed('accountId')) {
+      this.set('accountId', this.original['accountId']);
+    }
     super.discard();
```
sandikbarr: I found that BaseModel discard was clearing the accountId since it's a sort of pseudo field that comes from the parent account.

sandikbarr:
Thanks for the review @svevang :)
The bug on the series owner field was that discard didn't reset it to the original. You can only see the owner field if you have access to more than one account though (the staging admin user is set up for multiple account for example), so that could be why you're not seeing where the bug was.

svevang:
Got it now. Repro'd the bug and fix.  👍 looks great!

PR #642, opened 2019-05-08
Publish box refresh

pstjean:
A full featured publish box that owns the scheduling and validation functions of the episode creation flow.


src/app/story/directives/status.component.ts
```diff
@@ -1,64 +1,43 @@
...
+export type StoryStatus = 'draft' | 'scheduled' | 'published';
 @Component({
   selector: 'publish-story-status',
   styleUrls: ['status.component.css'],
   template: `
-    <h1>Publish</h1>
+    <h2 [class]="currentStatus">{{currentStatus}}</h2>
     <dl>
-
       <dt>Status</dt>
```
sandikbarr: I'm a little bothered by the semantics of these inputs not having labels in this `dl` `dt`/`dd`. Maybe Peter just kept it because it's what was there?

sandikbarr:
Not to be merged/deployed until Monday 6/17

src/app/story/directives/status-control.component.ts
```diff
@@ -0,0 +1,356 @@
...
+
+  saveAndPublish(toast: string) {
+    const wasNew = this.story.isNew;
+    this.story.save().subscribe(() => {
+      // a story should not publish while audio is processing
+      if (this.story.versions.every(version => version.files.every(file => !file.isProcessing))) {
+        this.togglePublish(toast);
+      } else {
+        const poll = interval(2000).pipe(
+          mergeMap(() => this.story.doc.reload()),
```
sandikbarr: Seems like I could just reload the files and not the whole story here.

src/app/story/directives/status.component.ts
```diff
@@ -1,64 +1,43 @@
...
-          <prx-button [model]="story" visible=1 orange=1 [disabled]="changed"
-            [working]="isPublishing" (click)="togglePublish()">Publish</prx-button>
-        </ng-container>
-        <ng-container *ngIf="isPublished && !strictInvalid">
-          <p *ngIf="changed">Unsaved changes</p>
-          <p *ngIf="!changed">Complete</p>
-        </ng-container>
+      <dt>Dropdate</dt>
+      <dd>
+        <prx-tz-datepicker *ngIf="nextStatus !== 'published' || currentStatus === 'published'; else publishImmediately"
```
cavis: So this is probably the only UI thing I feel strongly about in here... but what if we throw an "edit / change" toggle link on this datepicker?  Like the one that used to be next to the status:

image

It just feels a bit busy to always show these 3 fields, when you're most often not changing it.

src/app/story/directives/status.component.ts
```diff
@@ -1,64 +1,43 @@
...
+          [date]="date" (dateChange)="setDate($event)" [changed]="dateChanged">
+        </prx-tz-datepicker>
+        <ng-template #publishImmediately>Publish Immediately</ng-template>
```
cavis: +100 on the explicitness here.

src/app/story/directives/status-control.component.ts
```diff
@@ -0,0 +1,357 @@
...
+  color?: 'red' | 'orange' | 'green' | 'plain',
+  icon?: string
+}
+@Component({
+  selector: 'publish-status-control',
+  template: `
+    <hr>
+    <div class="actions" *ngIf="story">
+      <prx-button [model]="story" working=0 disabled=0 plain=1
+        [visible]="isChanged || nextStatus !== currentStatus" (click)="discard()">Discard</prx-button>
```
cavis: The alignment on this discard button irks me for some reason.  Would aligning it left help, or would that also be weird?  As in

image

src/app/story/directives/status.component.ts
cavis: Also found a bug here (which may be upstream)...

1. Publish an episode (doesn't happen when it's unpublished)
2. Change any field of the datepicker
3. Click "discard"
4. See the error: 
    ```
    ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'changed: false'. Current value: 'changed: true'.
    ```

I also notice that when you publish the episode, and change the field and change it back, it still has the "changed" outline on it.  (Probably related).

src/app/story/directives/status.component.ts
```diff
@@ -67,119 +46,73 @@ export class StoryStatusComponent implements DoCheck {
...
-        properties: { category: 'episode', label: 'episode/' + this.story.doc.id }});
-      this.toastr.success(`Episode ${this.story.publishedAt ? 'published' : 'unpublished'}`);
-      this.isPublishing = false;
-      this.editStatus = false;
-    });
+  notifyOfCanceledPublication() {
+    const futurePublished = this.story.publishedAt && new Date() < this.story.publishedAt;
+    const removingReleaseDate = this.story.changed('releasedAt') && !this.story.releasedAt;
+    if (removingReleaseDate && futurePublished) {
+      this.modal.alert(
```
cavis: This is very helpful.

src/app/story/directives/status-control.component.ts
```diff
@@ -0,0 +1,357 @@
...
+        visible="true"
+        [disabled]="buttons.primary.disabled"
+        [working]="working"
+        [red]="buttons.primary.color === 'red'"
+        [orange]="buttons.primary.color === 'orange'"
+        [green]="buttons.primary.color === 'green'"
+        [dropdown]="buttons.secondary"
+        (click)="buttons.primary.action()">
+        {{ buttons.primary.text }}
+        <div *ngIf="buttons.secondary" class="dropdown-menu-items">
```
cavis: Perhaps a question for @brandonhundt - but I find it odd that there are 2 dropdowns in this box.  The upper "status" one, which is only used until you schedule or publish.  And this lower one, which allows you to unpublish or delete.  Is that intentional to distinguish between the actions?

I guess I was expecting to see "Save and Publish Now" and "Save and Schedule" down in this dropdown (depending on if the date you selected is in the future or in the past):

image

src/app/story/directives/status-control.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    comp.story.invalid = (f, strict) => strict ? 'bad' : null;
+    fix.detectChanges();
+    expectDisabled(el, 'Found 1 Problem', false);
+    comp.story.invalid = (f, strict) => null;
+    comp.story.changed = () => true;
+    fix.detectChanges();
+    expectDisabled(el, 'Save & Publish', false);
+  });
+
+  cit('it correctly determines the next step', (fix, el, comp: StatusControlComponent) => {
```
cavis: Yeaaaah logic testing!

src/app/story/directives/status.component.css
```diff
@@ -4,65 +4,39 @@
...
-.status.draft {
+h2.draft {
   background-color: #ffa500;
 }
-.status.scheduled {
+h2.scheduled {
   background-color: #61A85D;
 }
-.status.published {
+h2.published {
```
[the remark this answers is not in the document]
brandonhundt: That's a fair point. If I had to choose we would change the published blue to a different color. Feel free to write a separate ticket for that as that would affect more than just this.

src/app/story/directives/status-control.component.ts
brandonhundt: Agreed with this one. Lets right align the discard button.

src/app/story/directives/status-control.component.ts
brandonhundt: I believe there was an issue with doing more than one action on save.

src/app/story/directives/status.component.ts
brandonhundt: Do you mean an edit button next to the dropdate field? If so, I would be okay with that.

src/app/story/directives/hero.component.ts
```diff
@@ -19,22 +19,12 @@ import { StoryModel } from '../../shared';
       </div>
       <div class="hero-info" *ngIf="story">
         <h2>{{story.title || '(Untitled)'}}</h2>
-        <p *ngIf="story?.isNew">Not saved</p>
-        <p *ngIf="!story?.isNew">Last saved at {{story.updatedAt | timeago}}</p>
-      </div>
-      <div class="hero-actions" *ngIf="story">
```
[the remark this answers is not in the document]
sandikbarr: Rick's actually working on that status bar right this moment with #658, so I saw that too and did leave that alone.

src/app/story/directives/status.component.ts
cavis: Yep, just to make things a little less "busy" when the episode is published.  (When you probably won't be editing the drop date).

src/app/story/directives/status-control.component.ts
cavis: Well, the API doesn't really support it.  But it seems like @sandikbarr has already worked around that limitation?  (If I edit the title, change draft -> published, and click "Publish Now" ... it both saves my title change and then publishes the episode).

sandikbarr: I worked thru the "Save and Publish" (dug thru the models and found what I needed to wait on after Save) so you can "Save & Publish" with a single button click. It takes a bit if your audio still needs to process, which is why I added the spinner to the Found 1 Problem button that shows while it needs to process.

src/app/story/directives/status.component.css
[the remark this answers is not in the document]
sandikbarr: Farski has that color palette sketch we might reference if we want to choose something else. The other blue option (other than tints and shades of this blue color) is `#17C9D3`.

src/app/story/directives/status.component.ts
sandikbarr: was just an accessibility nitpick, but labels also dont seem as useful in this box.

src/app/story/directives/status.component.css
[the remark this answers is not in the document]
brandonhundt: Just wrote this: https://github.com/PRX/publish.prx.org/issues/660

src/app/story/directives/status.component.ts
sandikbarr: I will look at that date discard bug. Good find.

I do feel like those three stacked fields make the box a little tall. I started to worry about screen resolution with the way that's fixed position.

src/app/story/directives/status.component.ts
sandikbarr: 😬 so this doesnt actually alert anymore because I found users can't remove the date (it's not valid and doesn't emit a change) but I kept it because _if_ things changed to where they could clear the date field and that was valid, we'd want to make sure they knew what would happen. It used to work with the old checkbox, so maybe we still need it given these new thoughts?

src/app/story/directives/status.component.ts
cavis: Good point.  Is there some chance this field would be unusable on a phone?

src/app/story/directives/status-control.component.ts
cavis: The AudioFileModels are already polling for changes, right?  (That's how the processing-progressbar gets data).  Would it be possible to just look at them, instead of reloading the story?

That said ... this is also fine.

src/app/story/directives/status.component.ts
cavis: Oh, sorry, I was thinking of the "You do not need to unpublish the episode to update the content" confirmation modal, not this one!

But I'm fine with leaving this in here, just in case we do get a datepicker field someday that allows clearing and emits null/blank.

src/app/story/directives/status.component.ts
sandikbarr: Fits on my iPhone X but we definitely dont have a lot of extra space in there.
IMG_0850

sandikbarr: Have to come up with a solution yet, but I at least see what happening a little better on this date discard bug. It's related to showing the publishedAt vs setting the releasedAt. The date picker emits a change after the discard which then sets releasedAt, when we were looking at publishedAt for published episodes, so that underlying swap seems to cause this error. So... working on it!

cavis: Would it be helpful to just set a `publishedReleasedAt` virtual field on the Model... that just "does the right thing" on encode/decode?  I guess it gets a bit complicated since publish/unpublish are standalone POST actions.

src/app/story/directives/status-control.component.ts
sandikbarr: you were absolutely correct on this, thank you! fixed

src/app/story/directives/hero.component.ts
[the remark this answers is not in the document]
sandikbarr: (nearly finished with it already and to be merged into this branch/PR)

src/app/story/directives/status.component.ts
sandikbarr: I fixed that date discard bug in the underlying datepicker in styleguide 2.1.12 https://github.com/PRX/styleguide.prx.org/pull/190 to not emit changes when the date value does not change. Seemed like we were getting extra emits there anyway.

src/app/story/directives/hero.component.ts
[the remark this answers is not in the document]
sandikbarr: merged! :tada:

src/app/story/directives/status.component.ts
sandikbarr: bolded but otherwise left alone

src/app/story/directives/hero.component.ts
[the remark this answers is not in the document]
cavis: +1 i dig it!

src/app/story/directives/status.component.ts
cavis: +1 works great now

cavis:
Okay, I think that resolves all the bugs.  My last UI nitpick was I wanted to hide the dropdate-pickers until you clicked some sort of "edit" link/toggle next to it.  Except on episodes with no date, where we want to encourage filling it in.

And the thing about the 2-different-dropdowns.

(But I'll defer to @brandonhundt on those).

sandikbarr:
I am styling up a date Edit toggle this morning, stay tuned!

I have made no changes for the dropdowns, but I see where Ryan is going with that.

sandikbarr:
Got that date toggle added. If we like it, I'll update the Distribution team on the change. It should affect the communication that went out yesterday, but they may want to mention it in the help article.

brandonhundt:
👍 :shipit: from me too!

PR #695, opened 2019-10-18
Add timepicker to series planner

pstjean:
Allows users to set the publish time for planned episodes.

### DIFF
--- a/src/app/series/directives/series-plan.component.html
+++ b/src/app/series/directives/series-plan.component.html
@@ -15,12 +15,10 @@
     publish once the state is changed to scheduled. This will prevent you from
     resetting the value on each individual episode.
   </p>
-  <div class="horizontal-checkboxes">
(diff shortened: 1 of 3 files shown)


src/app/series/directives/series-plan.component.html
```diff
@@ -8,6 +8,19 @@ <h1>Plan Episodes</h1>
   </p>
 </header>
 
+<section>
+  <h2>What time of day do you plan to publish your episodes?</h2>
```
cavis: `<feedback-not-demand-for-change>`

I don't love that this is the first section in the episode planner, or the description text below.  But I have no better suggestions (page is looking a bit crowded/busy in general, and words are hard).  So I'd defer to @brandonhundt on layout.

image

cavis:
Also, has codecov always been angry with Publish?

PR #726, opened 2020-09-24
Prevent accidental unpublish

### LINKED ISSUE #698: Guard against unpublish on stale state
Had an issue where a user may have accidentally unpublished from a stale browser window.  Unclear exactly how that may have happened, but for due diligence we should:

- [x] Refactor the story [StatusControlComponent](https://github.com/PRX/publish.prx.org/blob/master/src/app/story/directives/status-control.component.ts#L284) to guard against a stale state.  (Specifically, the `togglePublish` method and anything calling it)

cavis:
Closes #698 (again).

Belt and suspenders - an upstream bug in HalRemote fixed some token retrying.  But just in case that doesn't fix it ... also break up the `togglePublish()` method.  So if there's a stray observable somewhere, it doesn't accidentally toggle-to-unpublished.

### DIFF
--- a/src/app/story/directives/status-control.component.spec.ts
+++ b/src/app/story/directives/status-control.component.spec.ts
@@ -58,22 +58,22 @@ describe('StatusControlComponent', () => {
     expectDisabled(el, 'Save & Publish', true);
   });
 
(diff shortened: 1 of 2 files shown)


sandikbarr (approved):
that oughta do it

PR #739, opened 2021-01-25
Update upload component configurations

### LINKED ISSUE #730: Update Evaporate configuration
Based on some of the information I found while working on [these](https://github.com/PRX/internal/issues/420) [two](https://github.com/PRX/internal/issues/539) tickets, I think there are some changes that could be made to the env configuration used by Evaporate.

Currently, the relevant settings are:
```
USE_CLOUDFRONT=true
AWS_URL=https://dsn15xxxxxxx.cloudfront.net
BUCKET_ACCELERATION=true
```

Based on my reading of the Evaporate [docs](https://github.com/TTLabs/EvaporateJS/wiki/Evaporate.create()), the `cloudfront` and `s3Acceleration` options exist only to influence the automatic creation of the URL that it eventually tries to `OPTIONS`/`PUT`/`POST`/etc to.

With the current settings, it's generating URLS like: `https://prx-up.s3-accelerate.amazonaws.com/prod/8583b674…`

Evaporate is creating virtual host-style URLs because `s3Acceleration` is enabled. Having `cloudfront` set to true also would do that, but acceleration takes precedence. 

All that being said, there's no way to have it create IPv6-friendly URLs without explicitly providing an `aws_url` value. I think we should support that by using an `aws_url` like `https://prx-up.s3-accelerate.dualstack.amazonaws.com`. It sounds like even when you do this, you still need to keep `cloudfront` explicitly set to `true`, to [prevent the bucket name](https://github.com/TTLabs/EvaporateJS/blob/master/evaporate.js#L1037) from being added to the path.

The `s3Acceleration` option becomes unnecessary when using an explicit `aws_url`, as they are [mutually exclusive](https://github.com/TTLabs/EvaporateJS/blob/master/evaporate.js#L1942).

I would also suggest that the env vars be renamed to include some hint that they are applying to evaporate or uploads. `AWS_URL` and `USE_CLOUDFRONT` are quite ambiguous, especially when the app does other things with AWS and CloudFront.

I can't find anywhere in Publish or Styleguide (where a lot of evaporate functionality originates) where there current `USE_CLOUDFRONT` or `AWS_URL` values are even being used.

farski:
Closes #730

Blocked by https://github.com/PRX/styleguide.prx.org/pull/221

Updates components that use Styleguide upload service to use an explicit S3 endpoint URL, and drops the unused bucket acceleration flag.

Also updates related environment variables to be reduce ambiguity.

### DIFF
--- a/env-example
+++ b/env-example
@@ -19,11 +19,11 @@ AUTH_HOST=id.staging.prx.tech
 AUTH_CLIENT_ID=35zRqonwphKY05Uu2RA1YquwqyonSh3V6vAtnpB9
 
 # aws uploading
-BUCKET_NAME=prx-up
-BUCKET_FOLDER=dev
-BUCKET_ACCELERATION=true
-SIGN_URL=your-signing-url
-AWS_KEY=your-aws-key
+UPLOAD_S3_ENDPOINT_URL=https://publish-uploads-bucket.s3-accelerate.dualstack.amazonaws.com
+UPLOAD_BUCKET_NAME=publish-uploads-bucket
+UPLOAD_BUCKET_FOLDER=dev
+UPLOAD_SIGNING_SERVICE_URL=your-signing-url
+UPLOAD_SIGNING_SERVICE_KEY_ID=aws-key-used-by-signing-service
 
 # embeddable player
 PLAY_HOST=play.staging.prx.tech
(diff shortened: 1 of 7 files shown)


cavis (approved):
👍  looks good to me.  Will need to be deployed concurrent with some aws-secrets changes.

PR #753, opened 2021-08-02
Feat/alert on missing drafts

### LINKED ISSUE #749: Warn producers when the Production Calendar has no drafts
AdOps and Distro would like to alert producers when the production calendar has no drafts. 

Language

> ## Your production calendar is empty.
> The more drafts you add, the better we can support your podcast. Please add all of your known upcoming episodes to the Production Calendar.

Button: Proceed to Production Calendar 
Cancel

*If the podcast is marked as complete, the producer should not be warned.

Slack thread: https://prx.slack.com/archives/CG54MDMU6/p1621364722006700

svevang:
Closes #749 
This sets up a warning on the new episodes form, in the case where the producers drafts are running low.

### DIFF
--- a/src/app/story/directives/basic.component.ts
+++ b/src/app/story/directives/basic.component.ts
@@ -172,8 +172,7 @@ export class BasicComponent implements OnDestroy, DoCheck {
     const halParams = {
       page: 1,
       per: 1,
-      filters: 'state=unpublished',
-      sorts: 'released_at:desc',
+      filters: `after=${new Date().toISOString()}`,
       zoom: false
     };
 
--- a/src/app/story/directives/basic.component.ts
+++ b/src/app/story/directives/basic.component.ts
@@ -179,9 +179,9 @@ export class BasicComponent implements OnDestroy, DoCheck {
     let obs = this.series.doc.followItems('prx:stories', halParams);
 
     obs.subscribe((stories: HalDoc[]) => {
-      const story = stories[0] ? new StoryModel(this.story.parent, stories[0]) : null;
+      const story = stories[0] || null;
 
-      if (story && story.releasedAt > new Date) {
+      if (story && new Date(stories[0].releasedAt) > new Date) {
         this.hasDrafts = true;
       } else {
         this.hasDrafts = false;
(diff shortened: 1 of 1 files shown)


src/app/story/directives/basic.component.ts
```diff
@@ -108,8 +112,94 @@ export class BasicComponent implements OnDestroy, DoCheck {
...
+  }
+
+  loadUnpublished() {
+    this.hasDrafts = null;
+
+    const halParams = {
+      page: 1,
+      per: 1,
+      filters: 'state=unpublished',
+      sorts: 'released_at:desc',
```
cavis, line 176: Hit an edge case - apparently this sort puts null-released-at stories _before_ ones with a time set?

Digging around CMS filters for a way around that - noticed there's an `after` filter that wires into [this scope](https://github.com/PRX/cms.prx.org/blob/master/app/models/story.rb#L141).  Things work as expected if I drop the sort, and just do a:

```js
filters: `after=${new Date().toISOString()}`
```

Which actually might also save the trouble of initing the StoryModel at all.

cavis (changes requested):
Hit one edge case around null-released-at stories.  Think I found a CMS filter that may fix that.

But otherwise all looks/works good!

src/app/story/directives/basic.component.ts
```diff
@@ -108,8 +112,93 @@ export class BasicComponent implements OnDestroy, DoCheck {
...
+      page: 1,
+      per: 1,
+      filters: `after=${new Date().toISOString()}`,
+      zoom: false
+    };
+
+    let obs = this.series.doc.followItems('prx:stories', halParams);
+
+    obs.subscribe((stories: HalDoc[]) => {
+      const story = stories[0] ? new StoryModel(this.story.parent, stories[0]) : null;
```
cavis, line 182: One quirk of how HalDoc works ...

If you new a Model, it's by default going to [loadRelated](https://github.com/PRX/publish.prx.org/blob/master/src/app/shared/model/story.model.ts#L145).  Which for the StoryModel fires of 3 additional XHR requests (since you zoom=falsed, those resources weren't embedded).

But since the below check on line 184 is now redundant with the `after=` filter ... you could probably just forgo the StoryModel entirely here.

cavis (approved):
Does appear to run a few unnecessary GET requests, but hey, what's a bit of extra XHR?  All the things working!

+1!

PR #758, opened 2022-07-05
Multi feed wrap

### LINKED ISSUE #757: Multiple feeds feature wrap-up
Some fallout from #755.

- [x] Add field for "episode offset"
  - Just allow positive offsets for now, as negative ones are problematic for Dovetail Router
- [x] Add field for "keywords to include in feed"
  - Maybe reuse the story-tagging component?
  - Would be nice if it auto-populated from your episode tags ... but that may be too much work for now.
- [x] Add field for "audio encoding"
  - Needs to include/validate the format, bitrate/bits, channels, and frequency
  - Some way to just leave blank
  - When blank, make clear each episode will follow the 1st-original-segment's encoding
- ~Add button to duplicate a feed~
  - Skipped this one, as this is a very infrequent workflow. And optimizing for developer time!

ALSO:
- [x] Remove the top-bar external link for private feeds (the one with no token included)

cavis:
Fixes #757 

Adds 3 new fields to the Series -> RSS Feeds tab:

image

### DIFF
--- a/src/app/series/directives/series-feeds.component.html
+++ b/src/app/series/directives/series-feeds.component.html
@@ -157,6 +157,38 @@
             </div>
           </prx-fancy-field>
 
+          <prx-fancy-field label="Audio Format">
+            <div class="fancy-hint">
+              Optionally set an explicit audio format for your episodes. If you leave this blank, we will automatically use the same format
+              as the first audio segment you upload for each episode.
+            </div>
+            <div class="audio-format">
+              <prx-fancy-field [model]="f" name="audioType" [select]="audioTypeOptions"></prx-fancy-field>
+              <prx-fancy-field
+                required
+                [model]="f"
+                name="audioBitrate"
+                [select]="audioBitrateOptions"
+                *ngIf="f.audioType === 'mp3'"
+              ></prx-fancy-field>
+              <prx-fancy-field
+                required
+                [model]="f"
+                name="audioBitdepth"
+                [select]="audioBitdepthOptions"
+                *ngIf="f.audioType && f.audioType !== 'mp3'"
+              ></prx-fancy-field>
+              <prx-fancy-field
+                required
+                [model]="f"
+                name="audioChannel"
+                [select]="audioChannelOptions"
+                *ngIf="f.audioType"
+              ></prx-fancy-field>
+              <prx-fancy-field required [model]="f" name="audioSample" [select]="audioSampleOptions" *ngIf="f.audioType"></prx-fancy-field>
+            </div>
+          </prx-fancy-field>
+
           <button *ngIf="!f.isDefault" class="delete remove-feed" (click)="removeFeed(i)">
             <i class="remove-icon" aria-hidden="true"></i> Remove feed
           </button>
--- a/src/app/series/directives/series-feeds.component.html
+++ b/src/app/series/directives/series-feeds.component.html
@@ -157,13 +157,11 @@
             </div>
           </prx-fancy-field>
 
-          <prx-fancy-field
-            *ngIf="!f.isDefault"
-            label="Feed Episode Categories"
-            [model]="f"
-            name="includeTags"
-            [multiselect]="includeTagOptions"
-          >
+          <prx-fancy-field label="Episode Publish Delay" [model]="f" name="episodeOffsetSeconds" [select]="episodeOffsetOptions">
+            <div class="fancy-hint">Instead of episodes immediately showing up in this feed when published, delay their appearance.</div>
+          </prx-fancy-field>
+
+          <prx-fancy-field *ngIf="!f.isDefault" label="Episode Categories" [model]="f" name="includeTags" [multiselect]="includeTagOptions">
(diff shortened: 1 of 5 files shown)


src/app/series/directives/series-feeds.component.html
```diff
@@ -19,7 +19,7 @@ <h2>No feeds found! Please contact support</h2>
           <strong *ngIf="!f.isDefault">{{ f.title || '(New Feed)' }}</strong>
           <div *ngIf="!f.isNew" class="feed-url">
             {{ f.publishedFeedUrl() }}
-            <a [href]="f.publishedFeedUrl()" target="_blank" rel="noopener" class="external-icon"></a>
+            <a [href]="f.publishedFeedUrl(true)" target="_blank" rel="noopener" class="external-icon"></a>
```
cavis, line 22: Got some feedback that this link was confusing for private feeds.  Since it didn't include any `?auth=` token, you would open this link and get a 403 error.

This keeps the displayed value the same (no token) ... but the actual href points to your _last_ `?auth=` token if it's a private feed.

src/app/series/directives/series-feeds.component.ts
```diff
@@ -12,6 +12,49 @@ export class SeriesFeedsComponent implements OnDestroy {
...
   loading = true;
   expanded = [];
+  includeTagOptions: string[][];
+  episodeOffsetOptions = [
+    ['5 minutes', 300],
+    ['15 minutes', 900],
+    ['1 hour', 3600],
+    ['6 hours', 21600],
+    ['12 hours', 43200],
+    ['24 hours', 86400]
```
cavis, line 22: I just randomly picked some values here.  I'm actually not sure windowing will be useful to anyone yet?  Especially not until we support negative values - it's a bit confusing to set a positive offset on the default/public feed, and leave it blank on the private/subscriber feed.

src/app/series/directives/series-feeds.component.ts
```diff
@@ -12,6 +12,49 @@ export class SeriesFeedsComponent implements OnDestroy {
...
+  ];
+  audioSampleOptions = [
+    ['8 kHz', 8000],
+    ['11.025 kHz', 11025],
+    ['12 kHz', 12000],
+    ['16 kHz', 16000],
+    ['22.05 kHz', 22050],
+    ['24 kHz', 24000],
+    ['44.1 kHz', 44100],
+    ['48 kHz', 48000]
```
cavis, line 56: Copied all these values from the Feeder validations.

src/app/series/directives/series-feeds.component.ts
```diff
@@ -41,6 +84,11 @@ export class SeriesFeedsComponent implements OnDestroy {
         this.loading = false;
       }
     });
+    if (series.doc.has('prx:tags')) {
+      series.doc.follow('prx:tags').subscribe((tags) => {
+        this.includeTagOptions = (tags['tags'] || []).map((t) => [t, t]);
```
cavis, line 89: One bit of oddness:

Because this CMS endpoint does a mysql `SELECT DISTINCT`, you end up with a list of _case insensitive_ unique values here.  So if you tagged one episode with `hello` and another with `HeLLo` ... you'd only see 1 of those 2 values here.

BUT - as the Feeder logic for `include_tags` already does a bunch of normalization, things will work out as expected here.  And in fact, Feeder [goes further](https://github.com/PRX/feeder.prx.org/blob/master/app/models/feed.rb#L92) to normalize them, restricting characters.  So you might end up matching episodes you didn't expect.  Though in reality, I'll bet that doesn't happen.

src/app/shared/model/feeder-feed.model.spec.ts
```diff
@@ -72,6 +83,34 @@ describe('FeederFeedModel', () => {
     expect(feed.encode()['includeZones']).toBeNull();
   });
 
+  it('encodes included tags', () => {
+    feed.includeTags = ['anything'];
```
svevang, line 87: The include tag here is not a wildcard, but the 'anything' string literal.

svevang (approved):
Looks good!

src/app/series/directives/series-feeds.component.ts
```diff
@@ -12,6 +12,49 @@ export class SeriesFeedsComponent implements OnDestroy {
   podcast: FeederPodcastModel;
   loading = true;
   expanded = [];
+  includeTagOptions: string[][];
+  episodeOffsetOptions = [
+    ['5 minutes', 300],
```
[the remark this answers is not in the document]
cavis: Looks like it's running off `published_at` - and gets used in the [feed.feed_episodes](https://github.com/PRX/feeder.prx.org/blob/master/app/models/feed.rb#L77) logic to get the list of Episodes to write to the RSS.

But also, the `/api/v1/episodes/<guid>` endpoint DTR uses runs off the `episode.published_at`, but has no conception of which Feed you're using.  Which means we can't use negative values (yet) - any request to DTR for the episode will give you a 404 until we pass the real `published_at` date.

PR #765, opened 2022-08-30
Switch to AGPL-3.0

farski:


cavis:
Need to also change `package.json` reference to the license.

cavis (approved):
👍

PR #768, opened 2022-10-12
Series engagement settings

### LINKED ISSUE #766: Add payment pointer to the UI
Set the payment pointers for use in web monetization, that were added in feeder: https://github.com/PRX/feeder.prx.org/pull/429

cavis:
Fixes #766.  Adds an "Engagement Settings" tab, with the 2 fields we have already.  Also adds checkboxes to each feed to include/exclude them (defaults to true).

image

On the default feed:

And on non-default feeds:

### DIFF
--- a/src/app/series/directives/series-engagement.component.css
+++ b/src/app/series/directives/series-engagement.component.css
@@ -1,0 +1,6 @@
+/**
+ * Series engagement
+ */
+.engagement {
+  width: 100%;
+}
--- a/src/app/series/directives/series-engagement.component.html
+++ b/src/app/series/directives/series-engagement.component.html
@@ -18,6 +18,15 @@
       <div class="fancy-hint">
         The PRX Embeddable player support micropayments. We've created a guide to walk you through the process of collecting payments as
         folks listen.
+        <strong
+          ><a
+            target="_blank"
+            rel="noopener"
+            title="Setting Up Micropayments Wallet"
+            href="https://help.prx.org/hc/en-us/articles/9901810244251"
+            >Learn more about micropayments.</a
+          ></strong
+        >
       </div>
     </prx-fancy-field>
   </div>
--- a/src/app/series/directives/series-engagement.component.html
+++ b/src/app/series/directives/series-engagement.component.html
@@ -17,7 +17,7 @@
     <prx-fancy-field textinput [model]="podcast" name="paymentPointer" label="Micropayments Wallet">
       <div class="fancy-hint">
         The PRX Embeddable player support micropayments. We've created a guide to walk you through the process of collecting payments as
-        folks listen. <strong><a href="TODO">Learn more about micropayments.</a></strong>
+        folks listen.
       </div>
     </prx-fancy-field>
   </div>
--- a/src/app/series/directives/series-engagement.component.html
+++ b/src/app/series/directives/series-engagement.component.html
@@ -1,0 +1,24 @@
+<div *ngIf="!loading">
+  <prx-fancy-field label="Engagement Settings">
+    <div class="fancy-hint">Let your listeners know how they can interact with you further.</div>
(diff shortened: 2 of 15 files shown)


src/app/series/directives/series-engagement.component.html
```diff
@@ -0,0 +1,24 @@
...
+    <prx-fancy-field textinput [model]="podcast" name="donationUrl" label="Donation Form">
+      <div class="fancy-hint">
+        Let your listeners know where they can support your work by entering in an address to your donation form.
+      </div>
+    </prx-fancy-field>
+
+    <prx-fancy-field textinput [model]="podcast" name="paymentPointer" label="Micropayments Wallet">
+      <div class="fancy-hint">
+        The PRX Embeddable player support micropayments. We've created a guide to walk you through the process of collecting payments as
+        folks listen. <strong><a href="TODO">Learn more about micropayments.</a></strong>
```
cavis, line 20: @brandonhundt where should this link go, for now?  Or should I leave it out?

src/app/series/directives/series-feeds.component.html
```diff
@@ -230,6 +246,22 @@ <h2>No feeds found! Please contact support</h2>
             </div>
           </prx-fancy-field>
 
+          <prx-fancy-field label="Engagement Settings">
```
cavis, line 249: Note: there's a bunch of unrelated linting going on in this file.  (I didn't have Prettier setup correctly the last time I edited it).

But this is the new section of code.  Note that it shows up on _all_ feeds, even the default.

Question: is the correct place in the form for these checkboxes?

src/app/shared/model/feeder-podcast.model.ts
```diff
@@ -69,7 +73,9 @@ export class FeederPodcastModel extends BaseModel {
     enclosurePrefix: [URL('Not a valid URL')],
     summary: [LENGTH(0, 4000)],
     displayEpisodesCount: [POSITIVE_INT_OR_BLANK],
-    displayFullEpisodesCount: [POSITIVE_INT_OR_BLANK]
+    displayFullEpisodesCount: [POSITIVE_INT_OR_BLANK],
+    paymentPointer: [URL('Not a valid URL')],
+    donationUrl: [URL('Not a valid URL')]
```
cavis, line 78: Are both of these fields _always_ expected to be URLs?

src/app/series/directives/series-engagement.component.html
brandonhundt: Leave it out for now. I need to talk to distro about this. I'll start that convo.

src/app/shared/model/feeder-podcast.model.ts
brandonhundt: Yes. I think we can expect that. The payment pointer should be a link to the wallet.

brandonhundt (commented):
Noting the URL validation appears to prevent the give.prx.org URLs from working:

`https://give.prx.org/give/409824/#!/donation/checkout?c_src=Referral&c_src2=website-donate-heart`

^ this is The World's actual donation form. When I add the link, it says its an invalid link. This may prevent a lot of PRX donation links from being added.

Screen Shot 2022-10-13 at 1 49 12 PM

src/app/series/directives/series-engagement.component.css
```diff
@@ -0,0 +1,6 @@
+/**
+ * Series engagement
+ */
+.engagement {
+  width: 100%;
```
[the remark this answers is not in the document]
cavis: Because of the whole "shadow DOM" thing angular uses - yes.  And this file literally only applies to things directly in `series-engagement.component.html`.

(And we have no utility classes in Publish for things like this).

brandonhundt (approved):
Thanks for the fix! :shipit:

PR #776, opened 2023-09-28
Validate description bytes

cavis:
Validate the HTML-encoded description bytes is < 4000.

The flow goes like this:

1. User enters a description markdown of something like 3000 chars.  Frontend validations are fine with that.
2. User clicks save - CMS saves the markdown plus the HTML encoded version to the database.
3. After save, we decode the `description` field from the API (the HTML-encoded one)... calculate byte length, and highlight the field in red with the text "Description is too long"
4. Since this is a strict validation, user is prevented from publishing the episode until they fix the problem.

image

### DIFF
--- a/src/app/shared/model/story.model.ts
+++ b/src/app/shared/model/story.model.ts
@@ -36,7 +36,7 @@ const MIN = (min: number): BaseInvalid => {
 
 const MAXBYTES = (maxBytes: number): BaseInvalid => {
   return (key: string, value: any) => {
-    if (new TextEncoder().encode(value || '').length > maxBytes) {
+    if (typeof TextEncoder !== 'undefined' && new TextEncoder().encode(value || '').length > maxBytes) {
       return `${key} is too long`;
     } else {
       return null;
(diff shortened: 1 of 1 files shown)


src/app/shared/model/story.model.ts
```diff
@@ -160,10 +203,18 @@ export class StoryModel extends BaseModel implements HasUpload {
     this.updatedAt = new Date(this.doc['updatedAt']);
     this.publishedAt = this.doc['publishedAt'] ? new Date(this.doc['publishedAt']) : null;
     this.releasedAt = this.doc['releasedAt'] ? new Date(this.doc['releasedAt']) : null;
+
+    // count bytes of the html description
+    this.descriptionBytes = this.doc['descriptionMd'] || '';
```
svevang, line 208: Do we need this assignment if we're assigning the following if/else?

svevang (approved):
Looks good!  Just a question on assignment and branch coverage

src/app/shared/model/story.model.ts
cavis: Whoops!  Don't know why i left that in there.

